PR #8, opened 2022-10-12
initial deploy

agathanatasha:
# Things added
- setup kratos pebble container
- push configs and identity schema to container
- observe on install and pebble ready event

# To test
`juju deploy ./kratos_ubuntu-*-amd64.charm --trust --resource oci-image=oryd/kratos:v0.10.1`
The charm should be able to arrive at Active Status

Note: you would not be able to use the kratos api yet because the database connection is not setup properly. Will work on relation with postgres and migration in future PR

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -6,3 +6,12 @@ description: |
   Charmed Ory Kratos
 summary: |
   Identity and user management system
+containers:
+  kratos:
+    resource: oci-image
+resources:
+  oci-image:
+    type: oci-image
+    description: Kratos OCI oci-image
+    auto-fetch: true
+    upstream-source: oryd/kratos:v0.10.1
--- a/src/charm.py
+++ b/src/charm.py
@@ -6,10 +6,20 @@
 
 """A Juju charm for Ory Kratos."""
 
+import glob
 import logging
+from pathlib import Path
 
+from charmed_kubeflow_chisme.exceptions import ErrorWithStatus
+from charmed_kubeflow_chisme.kubernetes import KubernetesResourceHandler
+from charmed_kubeflow_chisme.pebble import update_layer
+from lightkube import Client
+from lightkube.core.exceptions import ApiError
+from lightkube.resources.apps_v1 import StatefulSet
 from ops.charm import CharmBase
 from ops.main import main
+from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
+from ops.pebble import Layer, PathError, ProtocolError
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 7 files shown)


src/manifests/svc.yaml
```diff
@@ -0,0 +1,46 @@
+---
+# Source: https://github.com/ory/k8s/k8s/helm/charts/kratos/templates/service-admin.yaml
```
natalian98, line 2: This link seems to have changed
```suggestion
# Source: https://github.com/ory/k8s/blob/master/helm/charts/kratos/templates/service-admin.yaml
```

src/manifests/svc.yaml
```diff
@@ -0,0 +1,46 @@
...
+  type: ClusterIP
+  ports:
+    - port: 80
+      targetPort: http-admin
+      protocol: TCP
+      name: http
+  selector:
+    app.kubernetes.io/name: {{ app_name }}
+---
+# Source: https://github.com/ory/k8s/k8s/helm/charts/kratos/templates/service-public.yaml
```
natalian98, line 25:
```suggestion
# Source: https://github.com/ory/k8s/blob/master/helm/charts/kratos/templates/service-public.yaml
```

src/manifests/svc.yaml
```diff
@@ -0,0 +1,46 @@
...
+# helm charm v0.25.3
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ app_name }}-admin
+  labels:
+    app.kubernetes.io/component: {{ app_name }}-admin
+    app.kubernetes.io/name: {{ app_name }}
+    app.kubernetes.io/managed-by: juju
+    app.juju.is/created-by: {{ app_name }}
```
natalian98, line 12: Is the presence of these labels supposed to help in resources removal by juju? They don't get cleaned up when the charm is removed. Maybe we need controller or model info in the metadata too? I will check this as I have same pb in hydra.

metadata.yaml
```diff
@@ -6,3 +6,12 @@ description: |
   Charmed Ory Kratos
 summary: |
   Identity and user management system
+containers:
+  kratos:
+    resource: oci-image
+resources:
+  oci-image:
+    type: oci-image
+    description: Kratos OCI oci-image
```
natalian98, line 15:
```suggestion
    description: Kratos oci-image
```

src/charm.py
```diff
@@ -19,6 +29,156 @@ class KratosCharm(CharmBase):
...
+        """Event Handler for install event.
+
+        - push configs
+        - apply service manifests
+        - update pod template to expose 2 ports
+        """
+        self.unit.status = MaintenanceStatus("Configuring/deploying resources")
+
+        if self.container.can_connect():
+            self._update_kratos_config()
```
natalian98, line 112: I think it's enough to update the config on pebble ready event

src/manifests/svc.yaml
agathanatasha:
```yaml
operator.juju.is/name=<app_name>   
operator.juju.is/target=application
```
These two are the important ones for removal. It only work on namespaced resources. 
These are covered (I think)
- Service
- Secret
- role binding, role
- custom resource / definition
- config map

src/manifests/svc.yaml
```diff
@@ -0,0 +1,46 @@
...
+    operator.juju.is/name: {{ app_name }}
+    operator.juju.is/target: application
+spec:
+  type: ClusterIP
+  ports:
+    - port: 80
+      targetPort: http-admin
```
nsklikas, line 19: You should be able to use `4434` instead of `http-admin` as `targetPort` in order to avoid having to patch the `StatefulSet`

src/manifests/svc.yaml
```diff
@@ -0,0 +1,46 @@
...
+    app.kubernetes.io/name: {{ app_name }}
+    app.kubernetes.io/managed-by: juju
+    app.juju.is/created-by: {{ app_name }}
+    operator.juju.is/name: {{ app_name }}
+    operator.juju.is/target: application
+spec:
+  type: ClusterIP
+  ports:
+    - port: 80
+      targetPort: http-public
```
nsklikas, line 42: Same here

src/manifests/svc.yaml
natalian98: Those services don't get deleted on remove event, but I guess that's a juju bug

src/manifests/svc.yaml
agathanatasha: I am not sure how it could avoid patching the StatefulSet. 4434 and 4433 are not exposed by default. I would still need to patch the statefulset when I use numbers instead of port name in the SVCs

nsklikas: I think it will work (I have tried this). According to the docs (https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#ports):
```
Not specifying a port here DOES NOT prevent that port from being exposed. Any port which is listening on the default "0.0.0.0" address inside a container will be accessible from the network
```

src/manifests/svc.yaml
agathanatasha: I added an on remove handler to remove them. Good catch

src/charm.py
```diff
@@ -19,6 +29,139 @@ class KratosCharm(CharmBase):
...
+
+        try:
+            self.resource_handler.apply()
```
nsklikas, line 116: Only the leader unit should apply the k8s manifests

src/charm.py
```diff
@@ -19,6 +29,139 @@ class KratosCharm(CharmBase):
...
+        self.unit.status = MaintenanceStatus("Configuring/deploying resources")
+
+        if self.container.can_connect():
+            self._update_kratos_config()
+            try:
+                update_layer(self._container_name, self.container, self._pebble_layer, logger)
+                self.unit.status = ActiveStatus()
+            except ErrorWithStatus as e:
+                self.model.unit.status = e.status
+                if isinstance(e.status, BlockedStatus):
```
nsklikas, line 146: Can `e.status` be anything other than `BlockedStatus`?

src/charm.py
```diff
@@ -19,6 +29,139 @@ class KratosCharm(CharmBase):
...
+                    logger.error(str(e.msg))
+                else:
+                    logger.info(str(e.msg))
+        else:
+            event.defer()
+            logger.info("Cannot connect to Kratos container. Deferring pebble ready event.")
+            self.unit.status = WaitingStatus("Waiting to connect to Kratos container")
+
+    def _on_remove(self, _) -> None:
+        """Event Handler for pebble ready event.
```
natalian98, line 156:
```suggestion
        """Event Handler for remove event.
```

src/charm.py
```diff
@@ -19,6 +29,139 @@ class KratosCharm(CharmBase):
...
+                }
+            },
+            "checks": {
+                "ready": {
+                    "override": "replace",
+                    "http": {"url": "http://localhost:4434/admin/health/ready"},
+                },
+                "alive": {
+                    "override": "replace",
+                    "http": {"url": "http://localhost:4434/admin/health/ready"},
```
natalian98, line 83: The `/health/alive` endpoint is for checking the http server status only, while `/health/ready` checks both the server and database connection. Maybe just the second one is enough? Anyway, they seem to be duplicated:
```suggestion
                    "http": {"url": "http://localhost:4434/admin/health/alive"},
```

src/manifests/svc.yaml
```diff
@@ -0,0 +1,46 @@
...
+    operator.juju.is/target: application
+spec:
+  type: ClusterIP
+  ports:
+    - port: 80
+      targetPort: 4433
+      protocol: TCP
+      name: http
+  selector:
+    app.kubernetes.io/name: {{ app_name }}
```
jnsgruk, line 46: I would drop this in favour of using the Kubernetes Service Patch lib if possible: https://charmhub.io/observability-libs/libraries/kubernetes_service_patch

PR #13, opened 2022-11-16
Working kratos charm

agathanatasha:
The PR contains all logic needed to have a functional kratos up and running
- apply k8s resources on install
- remove k8s resources on remove
- update container on database created
- update container and run sql migration on pebble ready


src/charm.py
```diff
@@ -19,6 +29,230 @@ class KratosCharm(CharmBase):
...
+        self._update_container(event)
+
+    def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
+        # Handle the created database
+        # Create configuration file for app
+        self.unit.status = MaintenanceStatus("Retrieving database details")
+
+        self._update_container(event)
+
+        self._run_sql_migration()
```
natalian98, line 242: I believe you don't need to run the migration here since it's already called in `_update_container`
```suggestion
```

src/charm.py
```diff
@@ -19,6 +24,181 @@ class KratosCharm(CharmBase):
...
+        if not self.model.relations["pg-database"]:
+            event.defer()
+            logger.error("Missing required relation with postgresql")
+            self.model.unit.status = BlockedStatus("Missing required relation with postgresql")
+            return
+
+        if not self.database.is_database_created():
+            event.defer()
+            logger.info("Missing database details. Deferring pebble ready event.")
+            self.unit.status = BlockedStatus("Waiting for database creation")
```
natalian98, line 164: How about `WaitingStatus` instead? `BlockedStatus` suggests that a manual intervention is required.

src/charm.py
```diff
@@ -19,6 +24,181 @@ class KratosCharm(CharmBase):
...
+        try:
+            self._update_layer()
+        except ChangeError as err:
+            logger.error(str(err))
+            self.unit.status = BlockedStatus("Failed to replan")
+            return
+
+        self._run_sql_migration()
+
+        self.unit.status = ActiveStatus()
```
natalian98, line 178: What do you think about setting the status in `_on_pebble_ready` and `_on_database_created` events instead?

src/charm.py
agathanatasha: Good catch, will update that

src/charm.py
agathanatasha: I want to, but I am already updating statuses when checking for container and database connection. I thought it would be more consistent to set all statuses in this method. I am open to suggestions

src/charm.py
```diff
@@ -19,6 +24,181 @@ class KratosCharm(CharmBase):
...
+            self.unit.status = MaintenanceStatus("Applying new pebble layer")
+            self._container.add_layer(self._container_name, new_layer, combine=True)
+            logger.info("Pebble plan updated with new configuration, replanning")
+            self._container.replan()
+
+        # Compare changes in kratos config
+        try:
+            current_config = self._container.pull(self._config_file_path).read()
+        except PathError:
+            self._container.push(self._config_file_path, self._config, make_dirs=True)
```
natalian98, line 116: To avoid the charm getting into error state when the directory doesn't exist yet, we could push the config file while comparing the layers.
They will differ the first time they're compared, and we can assume that the config will not exist neither. We could force-push it once and then compare against the current state.

src/charm.py
```diff
@@ -19,6 +24,181 @@ class KratosCharm(CharmBase):
...
+                schema = schema_file.read()
+                self._container.push(self._identity_schema_file_path, schema, make_dirs=True)
+            logger.info("Pushed configs to kratos container")
+        except (ProtocolError, PathError) as e:
+            logger.error(str(e))
+            self.unit.status = BlockedStatus(str(e))
+
+    def _update_container(self, event) -> None:
+        """Update configs, pebble layer and run database migration."""
+        if not self._container.can_connect():
```
natalian98, line 149: I think we should also check the leadership

src/charm.py
```diff
@@ -19,6 +24,181 @@ class KratosCharm(CharmBase):
...
+
+        self.database = DatabaseRequires(
+            self,
+            relation_name="pg-database",
+            database_name="database",
+            extra_user_roles="SUPERUSER",
+        )
+
+        self.framework.observe(self.on.kratos_pebble_ready, self._on_pebble_ready)
+        self.framework.observe(self.database.on.database_created, self._on_database_created)
```
natalian98, line 42: How about observing `endpoints_changed` as well? They _shouldn't_ change too often, but we could have it just in case

src/charm.py
```diff
@@ -19,6 +24,181 @@ class KratosCharm(CharmBase):
...
+            logger.info("Pushed configs to kratos container")
+        except (ProtocolError, PathError) as e:
+            logger.error(str(e))
+            self.unit.status = BlockedStatus(str(e))
+
+    def _update_container(self, event) -> None:
+        """Update configs, pebble layer and run database migration."""
+        if not self._container.can_connect():
+            event.defer()
+            logger.info("Cannot connect to Kratos container. Deferring pebble ready event.")
```
natalian98, line 151: I think we're deferring an event in general - it could be either pebble ready or database created, since they both use this helper

src/charm.py
```diff
@@ -19,6 +24,183 @@ class KratosCharm(CharmBase):
...
+        current_layer = self._container.get_plan()
+        # Create a new config layer
+        new_layer = self._pebble_layer
+
+        if current_layer.services != new_layer.services:
+            self.unit.status = MaintenanceStatus("Applying new pebble layer")
+            self._container.push(self._config_file_path, self._config, make_dirs=True)
+            self._container.add_layer(self._container_name, new_layer, combine=True)
+            logger.info("Pebble plan updated with new configuration, replanning")
+            self._container.replan()
```
nsklikas, line 112: I have seen that in other charms this is written like this:

```python
        self._container.add_layer(self.container_name, new_layer, combine=True)
        self._container.autostart()
```

Is there a difference between the two? Do we need to check for diffs in the layers?

src/charm.py
```diff
@@ -19,6 +24,183 @@ class KratosCharm(CharmBase):
...
+        self._set_default_identity_schema()
+
+        try:
+            self._update_layer()
+        except ChangeError as err:
+            logger.error(str(err))
+            self.unit.status = BlockedStatus("Failed to replan")
+            return
+
+        self._run_sql_migration()
```
nsklikas, line 178: What happens if it is a breaking migration? Do we want to always auto run the migration? The ory page urges the user to take a backup before running migrations. Should we warn the user? 

This comment is probably out of scope for this PR.

src/charm.py
```diff
@@ -19,6 +24,183 @@ class KratosCharm(CharmBase):
...
+
+    def _run_sql_migration(self) -> None:
+        """Runs database migration."""
+        process = self._container.exec(
+            ["kratos", "migrate", "sql", "-e", "--config", self._config_file_path, "--yes"],
+            timeout=20.0,
+        )
+        try:
+            stdout, _ = process.wait_output()
+            logger.info(f"Executing automigration: {stdout}")
```
nsklikas, line 190:
```suggestion
            stdout, _ = process.wait_output()
            logger.info(f"Successfully executed automigration: {stdout}")
```

tests/unit/test_charm.py
```diff
@@ -1,15 +1,184 @@
...
+    harness.add_relation_unit(db_relation_id, "postgresql-k8s/0")
+    harness.update_relation_data(
+        db_relation_id,
+        "postgresql-k8s",
+        {
+            "data": '{"database": "database", "extra-user-roles": "SUPERUSER"}',
+            "endpoints": DB_ENDPOINTS,
+            "password": DB_PASSWORD,
+            "username": DB_USERNAME,
+        },
```
nsklikas, line 75: Since this is used in several tests, you could move it into a fixture.

nsklikas (changes requested):

src/charm.py
```diff
@@ -19,6 +24,183 @@ class KratosCharm(CharmBase):
...
+            return
+
+        self._run_sql_migration()
+
+        self.unit.status = ActiveStatus()
+
+    def _run_sql_migration(self) -> None:
+        """Runs database migration."""
+        process = self._container.exec(
+            ["kratos", "migrate", "sql", "-e", "--config", self._config_file_path, "--yes"],
```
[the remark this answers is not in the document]
natalian98: I think it still needs the `-e` flag to be able to read from the config file, see [here](https://www.ory.sh/docs/hydra/cli/hydra-migrate-sql).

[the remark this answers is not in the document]
agathanatasha: I agree this is a strange one, but I do need `-e`. I tried running without, it didn't work. This is also the same command in the job generated from helm chart

src/charm.py
agathanatasha: Definitely one of my worries too. I read on ory slack awhile back that it is safe to rerun migration. That's why I put this here for now. I have no idea how upgrades will be handled. Now that I am thinking about it, I don't think the set of migration would change since the kratos cli is tied to the image version. I am going to look into this further soon and tackle that in a separate PR.

src/charm.py
```diff
@@ -19,6 +24,183 @@ class KratosCharm(CharmBase):
...
+            logger.error(str(e))
+            self.unit.status = BlockedStatus(str(e))
+
+    def _update_container(self, event) -> None:
+        """Update configs, pebble layer and run database migration."""
+        if not self.unit.is_leader():
+            event.defer()
+            logger.info("Does not have leadership. Deferring event.")
+            self.unit.status = WaitingStatus("Waiting for leadership")
+            return
```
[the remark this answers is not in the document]
agathanatasha: good point, I will move the condition near the migration run for now. I will address all the migration concern in a separate PR

src/charm.py
agathanatasha: I honestly don't know. I posted a question on the public mattermost.

src/charm.py
```diff
@@ -19,6 +24,169 @@ class KratosCharm(CharmBase):
...
+
+        if not self.unit.is_leader():
+            event.defer()
+            logger.info("Does not have leadership. Deferring event.")
+            self.unit.status = WaitingStatus("Waiting for leadership")
+            return
+
+        self._run_sql_migration()
+
+        self.unit.status = ActiveStatus()
```
nsklikas, line 166: It feels like too many stuff is happening in this (+`_update_layer`) method. It is handling pebble readiness, changes in configuration and changes in the database. I think that it's better to break this into smaller functions, as we don't need to be doing all of those things on every event. E.g.:
- On pebble ready: update/push the config, push the layer, update status. This will probably raise errors as there will be no db, but we will be in a `BlockedStatus` so it's expected. We could defer the pebble ready event until the database is there, but IMHO that would complicate the flow without making the UX much better.
- On config changed: update/push the config, update status
- On database created: run migration, update/push the config, update status
- On database endpoint changed: update/push the config, update status

What do you think?

metadata.yaml
```diff
@@ -6,3 +6,16 @@ description: |
...
 summary: |
   Identity and user management system
+containers:
+  kratos:
+    resource: oci-image
+resources:
+  oci-image:
+    type: oci-image
+    description: Kratos oci-image
+    auto-fetch: true
```
nsklikas, line 16: What is the use of this param? I can't find any documentation on it, only some reactive charms that are using it.

src/charm.py
```diff
@@ -19,6 +24,169 @@ class KratosCharm(CharmBase):
...
+
+    def _update_layer(self) -> None:
+        """Updates the Pebble configuration layer and kratos config if changed."""
+        if not self._container.get_plan().to_dict():
+            self.unit.status = MaintenanceStatus("Applying new pebble layer")
+            self._container.push(self._config_file_path, self._config, make_dirs=True)
+            with open("src/identity.default.schema.json", encoding="utf-8") as schema_file:
+                schema = schema_file.read()
+                self._container.push(self._identity_schema_file_path, schema, make_dirs=True)
+            self._container.add_layer(self._container_name, self._pebble_layer, combine=True)
```
natalian98, line 108: You are adding the layer only once. It makes sense as long as there are no configurable options, but once we add any, the layer will need to be added and container replanned not only when the plan is empty.

src/charm.py
agathanatasha: The problem is that the order of events is not guaranteed. On database created could happen before on pebble ready. I am not observing config changed right now, but it could also happen before on pebble ready. I don't know for sure the state of the charm when an event is emitted. That's why I am coding in a very defensive manner. 
Regarding migration, this is an interim solution. I will open a separate PR for better handling migration.

src/charm.py
agathanatasha: Right now the layer doesn't change at all. We could update this in the future if the layer changes

src/charm.py
nsklikas: Still, if on every event you are forcing idempotent changes to the state of the container, you don't have to worry about the order of the events. For example if database_created comes before pebble_ready shouldn't matter as:
- The database_created hook will try migrating the db (it will succeed), updating the config (this will return without doing any work as the container will not be ready), restarting the service (this will return without doing any work as the container will not be ready), updating the status (this will set the status to maintenance or something).
- The pebble_ready hook will push the config (which will include the database config), push the pebble layer, update the status (this will set the status to active as both the container and the db will be ready).

src/charm.py
```diff
@@ -19,6 +24,165 @@ class KratosCharm(CharmBase):
...
+            logger.info("Pebble plan updated with new configuration, replanning")
+            self._container.replan()
+        else:
+            # Compare changes in kratos config
+            current_config = self._container.pull(self._config_file_path).read()
+            if current_config != self._config:
+                self.unit.status = MaintenanceStatus("Updating Kratos Config")
+                self._container.push(self._config_file_path, self._config, make_dirs=True)
+                logger.info("Updated kratos config")
+                self._container.restart()
```
nsklikas, line 118:
```suggestion
                self._container.restart(self._container_name)
```

nsklikas (approved):
LGTM

PR #17, opened 2022-12-01
Restructure data migration

agathanatasha:
Only run database migration on database created event
Updated unit tests

### DIFF
--- a/.github/workflows/on_pull_request.yaml
+++ b/.github/workflows/on_pull_request.yaml
@@ -17,7 +17,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Install dependencies
-        run: python3 -m pip install tox==3.27.1
+        run: python3 -m pip install tox
 
       - name: Run linters
         run: tox -e lint
--- a/.github/workflows/on_pull_request.yaml
+++ b/.github/workflows/on_pull_request.yaml
@@ -17,7 +17,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Install dependencies
-        run: python3 -m pip install tox
+        run: python3 -m pip install tox==3.27.1
 
       - name: Run linters
         run: tox -e lint
--- a/src/charm.py
+++ b/src/charm.py
@@ -103,9 +103,6 @@ class KratosCharm(CharmBase):
     def _get_database_relation_info(self) -> dict:
         """Get database info from relation data bag."""
         relation_id = self.database.relations[0].id
-        print(
-            f"DEBUGGING ~ file: charm.py ~ line 103 ~ self.database.relations: {self.database.relations}"
-        )
         relation_data = self.database.fetch_relation_data()[relation_id]
 
         return {
--- a/src/charm.py
+++ b/src/charm.py
@@ -129,12 +129,16 @@ class KratosCharm(CharmBase):
             return
 
         self.unit.status = MaintenanceStatus("Configuring/deploying resources")
+
         with open("src/identity.default.schema.json", encoding="utf-8") as schema_file:
             schema = schema_file.read()
             self._container.push(self._identity_schema_file_path, schema, make_dirs=True)
         self._container.add_layer(self._container_name, self._pebble_layer, combine=True)
         logger.info("Pebble plan updated with new configuration, replanning")
         self._container.replan()
+        if self.database.is_database_created():
+            self._container.push(self._config_file_path, self._config, make_dirs=True)
+            self._container.start(self._container_name)
         self.unit.status = ActiveStatus()
 
     def _on_database_created(self, event) -> None:
--- a/src/charm.py
+++ b/src/charm.py
@@ -70,7 +70,10 @@ class KratosCharm(CharmBase):
 
     @property
     def _config(self) -> str:
-        db_info = self._get_database_relation_info()
+        try:
+            db_info = self._get_database_relation_info() or {}
+        except IndexError:
+            db_info = {}
         config = {
             "log": {"level": "trace"},
             "identity": {
@@ -100,6 +103,9 @@ class KratosCharm(CharmBase):
     def _get_database_relation_info(self) -> dict:
         """Get database info from relation data bag."""
         relation_id = self.database.relations[0].id
+        print(
+            f"DEBUGGING ~ file: charm.py ~ line 103 ~ self.database.relations: {self.database.relations}"
+        )
         relation_data = self.database.fetch_relation_data()[relation_id]
 
         return {
@@ -138,13 +144,13 @@ class KratosCharm(CharmBase):
         with open("src/identity.default.schema.json", encoding="utf-8") as schema_file:
             schema = schema_file.read()
             self._container.push(self._identity_schema_file_path, schema, make_dirs=True)
+        self._container.push(self._config_file_path, self._config, make_dirs=True)
         self._container.add_layer(self._container_name, self._pebble_layer, combine=True)
         logger.info("Pebble plan updated with new configuration, replanning")
         self._container.replan()
 
         # in case container was terminated unexpectedly
         if self.database.is_database_created():
-            self._container.push(self._config_file_path, self._config, make_dirs=True)
             self._container.start(self._container_name)
             self.unit.status = ActiveStatus()
             return
--- a/src/charm.py
+++ b/src/charm.py
@@ -108,17 +108,22 @@ class KratosCharm(CharmBase):
             "endpoints": relation_data.get("endpoints"),
         }
 
-    def _run_sql_migration(self) -> None:
-        """Runs database migration."""
-        process = self._container.exec(
-            ["kratos", "migrate", "sql", "-e", "--config", self._config_file_path, "--yes"],
-        )
+    def _run_sql_migration(self) -> bool:
+        """Runs database migration.
+
+        Returns True if migration was run successfully, else returns false.
+        """
         try:
-            stdout, _ = process.wait_output()
-            logger.info(f"Successfully executed automigration: {stdout}")
+            process = self._container.exec(
+                ["kratos", "migrate", "sql", "-e", "--config", self._config_file_path, "--yes"],
+            )
         except ExecError as err:
             logger.error(f"Exited with code {err.exit_code}. Stderr: {err.stderr}")
-            self.unit.status = BlockedStatus("Database migration job failed")
+            return False
+
+        stdout, _ = process.wait_output()
+        logger.info(f"Successfully executed automigration: {stdout}")
+        return True
 
     def _on_pebble_ready(self, event) -> None:
         """Event Handler for pebble ready event."""
@@ -136,10 +141,18 @@ class KratosCharm(CharmBase):
         self._container.add_layer(self._container_name, self._pebble_layer, combine=True)
         logger.info("Pebble plan updated with new configuration, replanning")
         self._container.replan()
+
+        # in case container was terminated unexpectedly
         if self.database.is_database_created():
             self._container.push(self._config_file_path, self._config, make_dirs=True)
             self._container.start(self._container_name)
-        self.unit.status = ActiveStatus()
+            self.unit.status = ActiveStatus()
+            return
+
+        if self.model.relations["pg-database"]:
+            self.unit.status = WaitingStatus("Waiting for database creation")
+        else:
+            self.unit.status = BlockedStatus("Missing postgres database relation")
 
     def _on_database_created(self, event) -> None:
         """Event Handler for database created event."""
@@ -158,15 +171,20 @@ class KratosCharm(CharmBase):
 
         try:
             self._container.get_service(self._container_name)
-            logger.info("Updating Kratos config and restarting service")
-            self._container.push(self._config_file_path, self._config, make_dirs=True)
-            self._run_sql_migration()
-            self._container.start(self._container_name)
-            self.unit.status = ActiveStatus()
         except (ModelError, RuntimeError):
             event.defer()
             self.unit.status = WaitingStatus("Waiting for Kratos service")
             logger.info("Kratos service is absent. Deferring database created event.")
+            return
+
+        logger.info("Updating Kratos config and restarting service")
+        self._container.push(self._config_file_path, self._config, make_dirs=True)
+        is_migration_success = self._run_sql_migration()
+        if is_migration_success:
+            self._container.start(self._container_name)
+            self.unit.status = ActiveStatus()
+        else:
+            self.unit.status = BlockedStatus("Database migration failed.")
 
     def _on_database_changed(self, event: DatabaseCreatedEvent) -> None:
         """Event Handler for database changed event."""
@@ -180,13 +198,15 @@ class KratosCharm(CharmBase):
 
         try:
             self._container.get_service(self._container_name)
-            self._container.push(self._config_file_path, self._config, make_dirs=True)
-            self._container.restart(self._container_name)
-            self.unit.status = ActiveStatus()
         except (ModelError, RuntimeError):
             event.defer()
             self.unit.status = WaitingStatus("Waiting for Kratos service")
             logger.info("Kratos service is absent. Deferring database created event.")
+            return
+
+        self._container.push(self._config_file_path, self._config, make_dirs=True)
+        self._container.restart(self._container_name)
+        self.unit.status = ActiveStatus()
 
 
 if __name__ == "__main__":
--- a/src/charm.py
+++ b/src/charm.py
@@ -13,8 +13,8 @@ from charms.data_platform_libs.v0.database_requires import DatabaseCreatedEvent,
 from charms.observability_libs.v0.kubernetes_service_patch import KubernetesServicePatch
 from ops.charm import CharmBase
 from ops.main import main
-from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
-from ops.pebble import ChangeError, ExecError, Layer
+from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, ModelError, WaitingStatus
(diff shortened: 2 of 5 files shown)


src/charm.py
```diff
@@ -128,61 +108,79 @@ def _get_database_relation_info(self) -> dict:
...
+        with open("src/identity.default.schema.json", encoding="utf-8") as schema_file:
+            schema = schema_file.read()
+            self._container.push(self._identity_schema_file_path, schema, make_dirs=True)
+        self._container.add_layer(self._container_name, self._pebble_layer, combine=True)
+        logger.info("Pebble plan updated with new configuration, replanning")
+        self._container.replan()
+        self.unit.status = ActiveStatus()
 
-        if not self.database.is_database_created():
+    def _on_database_created(self, event) -> None:
```
natalian98, line 140: You observe `database_created` event, but run `_on_database_changed` method when it's emitted. Did you mean to run this one instead? It's not used anywhere else

src/charm.py
```diff
@@ -128,61 +108,79 @@ def _get_database_relation_info(self) -> dict:
...
+
         try:
-            self._update_layer()
-        except ChangeError as err:
-            logger.error(str(err))
-            self.unit.status = BlockedStatus("Failed to replan")
+            self._container.get_service(self._container_name)
+            logger.info("Updating Kratos config and restarting service")
+            self._container.push(self._config_file_path, self._config, make_dirs=True)
+            self._run_sql_migration()
```
natalian98, line 154: As mentioned by @nsklikas, the migration should be run only by leader unit

src/charm.py
```diff
@@ -128,61 +108,79 @@ def _get_database_relation_info(self) -> dict:
...
+        """Event Handler for database created event."""
+        if not self._container.can_connect():
             event.defer()
-            logger.info("Missing database details. Deferring event.")
-            self.unit.status = WaitingStatus("Waiting for database creation")
+            logger.info("Cannot connect to Kratos container. Deferring event.")
+            self.unit.status = WaitingStatus("Waiting to connect to Kratos container")
             return
 
+        self.unit.status = MaintenanceStatus("Creating database.")
```
natalian98, line 148: Database is already created at that point, how about a message about configuring the container/resources?

natalian98:
If the pod or container get restarted, the config file is not pushed and the kratos service is unaccessible.
To reproduce:
On a fresh deployment:
```
$ curl http://10.152.183.11:4433/schemas
[{"id":"default","schema":{"$id":"https://schemas.ory.sh/presets/kratos/identity.email.schema.json","$schema":"http://json-schema.org/draft-07/schema#","title":"Person","type":"object","properties":{"traits":{"type":"object","properties":{"email":{"type":"string","format":"email","title":"E-Mail","ory.sh/kratos":{"credentials":{"password":{"identifier":true}},"recovery":{"via":"email"},"verification":{"via":"email"}}}},"required":["email"],"additionalProperties":false}}}}]
```
After pod or container restart:
`kubectl exec -it kratos-0 -c kratos -n kratos -- sh -c "kill 1"`
Kratos gets recreated and the charm is active, but since the file is not pushed, its API and CLI are unaccessible:
```
$ curl http://10.152.183.11:4433/schemas
curl: (7) Failed to connect to 10.152.183.11 port 4433 after 0 ms: Connection refused
```
```
/home/ory # kratos list identities --endpoint http://127.0.0.1:4434/
2022/12/06 11:50:27 [DEBUG] GET http://127.0.0.1:4434//admin/identities
2022/12/06 11:50:27 [ERR] GET http://127.0.0.1:4434//admin/identities request failed: Get "http://127.0.0.1:4434//admin/identities": dial tcp 127.0.0.1:4434: connect: connection refused
2022/12/06 11:50:27 [DEBUG] GET http://127.0.0.1:4434//admin/identities: retrying in 1s (4 left)
2022/12/06 11:50:28 [ERR] GET http://127.0.0.1:4434//admin/identities request failed: Get "http://127.0.0.1:4434//admin/identities": dial tcp 127.0.0.1:4434: connect: connection refused
2022/12/06 11:50:28 [DEBUG] GET http://127.0.0.1:4434//admin/identities: retrying in 2s (3 left)
```
The health and readiness checks fail as well.

src/charm.py
```diff
@@ -97,75 +97,21 @@ def _config(self) -> str:
...
 
-    def _update_container(self, event) -> None:
-        """Update configs, pebble layer and run database migration."""
-        if not self._container.can_connect():
-            event.defer()
-            logger.info("Cannot connect to Kratos container. Deferring event.")
-            self.unit.status = WaitingStatus("Waiting to connect to Kratos container")
-            return
-
-        if not self.model.relations["pg-database"]:
```
natalian98, line 139: Don't we need to still make sure that the relation is in place?

nsklikas (changes requested):
- I seems like `_on_database_created` and `_on_database_changed` share a lot of code and their only difference is that `_on_database_created` also runs the migration. We could try to refactor them to reduce code duplication.
- We could handle the units status in an `update_status` hook. It would do all the necessary checks (`get_service`, `can_connect`, `is_database_created`) and set the corresponding status. hooks can call this to set the unit's status. What do you think about this?
- What happens on upgrade? Would the migration run in that case?

src/charm.py
```diff
@@ -39,7 +39,7 @@ def __init__(self, *args):
         )
 
         self.framework.observe(self.on.kratos_pebble_ready, self._on_pebble_ready)
-        self.framework.observe(self.database.on.database_created, self._on_database_changed)
+        self.framework.observe(self.database.on.database_created, self._on_database_created)
         self.framework.observe(self.database.on.endpoints_changed, self._on_database_changed)
```
natalian98, line 43: I know it was me who suggested observing `endpoints_changed`, but after giving it a second thought, we may not need it. I initially thought the endpoint would be an ip and thus could change, but it's referencing the postgresql-k8s-primary's svc name, so it will always remain `postgresql-k8s-primary.namespace.svc.cluster.local:5432`. What do you think?

src/charm.py
```diff
@@ -176,13 +122,71 @@ def _run_sql_migration(self) -> None:
...
+            self.unit.status = WaitingStatus("Waiting to connect to Kratos container")
+            return
+
+        self.unit.status = MaintenanceStatus("Updating database details")
+
+        try:
+            self._container.get_service(self._container_name)
+            self._container.push(self._config_file_path, self._config, make_dirs=True)
+            self._container.restart(self._container_name)
+            self.unit.status = ActiveStatus()
```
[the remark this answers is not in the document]
agathanatasha: I only want to run restart if the service is there

src/charm.py
```diff
@@ -176,13 +122,71 @@ def _run_sql_migration(self) -> None:
...
-    def _on_database_changed(self, event: DatabaseCreatedEvent) -> None:
+        with open("src/identity.default.schema.json", encoding="utf-8") as schema_file:
+            schema = schema_file.read()
+            self._container.push(self._identity_schema_file_path, schema, make_dirs=True)
+        self._container.add_layer(self._container_name, self._pebble_layer, combine=True)
+        logger.info("Pebble plan updated with new configuration, replanning")
+        self._container.replan()
+        if self.database.is_database_created():
+            self._container.push(self._config_file_path, self._config, make_dirs=True)
+            self._container.start(self._container_name)
```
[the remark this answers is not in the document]
agathanatasha: I set the service to startup disabled because it depends on a valid config file which might not be available at the time of the event. This snippet is added to make sure the container can recover if it was terminated unexpectedly (see earlier comments from Natalia)

src/charm.py
```diff
@@ -51,7 +51,7 @@ def _pebble_layer(self) -> Layer:
                 self._container_name: {
                     "override": "replace",
                     "summary": "Kratos Operator layer",
-                    "startup": "enabled",
+                    "startup": "disabled",
```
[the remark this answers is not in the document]
agathanatasha: I want to only start the service when we have all the config

agathanatasha:
> * I seems like `_on_database_created` and `_on_database_changed` share a lot of code and their only difference is that `_on_database_created` also runs the migration. We could try to refactor them to reduce code duplication.
> 
>     * We could handle the units status in an `update_status` hook. It would do all the necessary checks (`get_service`, `can_connect`, `is_database_created`) and set the corresponding status. hooks can call this to set the unit's status. What do you think about this?
> 
>     * What happens on upgrade? Would the migration run in that case?

Right now, migration is not ran on upgrade. Data platform team's recommendation is to have the user manually trigger an upgrade. So we would add a run database migration action when we work on updating Kratos versions.

I think on database changed would have more common with config change (which is the next thing to work on). On database created would have more in common with a run db migration action. With that in mind, I think we should keep it separate.

Update status hook is run every 5 mins and often longer on real customers' deployment. I don't think it's a good idea to rely on it for setting statuses.

tests/unit/test_charm.py
```diff
@@ -25,9 +25,20 @@ def setup_postgres_relation(harness):
     )
```
gruyaume, line 26: A note on mocking, we don't want to mock the method `mocked_sql_migration`. What we really want to mock is the 'container.exec` command in it. The reason is that if we end up refactoring the charm code one day and change the place where the sql migrations are run, now all the unit tests fail even though the behaviour may have remained the same. If the code is refactored but the behaviour remained the same, unit tests should still pass.

gruyaume: The same is true of `mocked_update_container`

gruyaume: We should aim at removing those mocks. This may end up in our tests being too complex. This is good because it'll reflect the fact that the charms are also too complex and puts pressure for a better design

tests/unit/test_charm.py
```diff
@@ -25,9 +25,20 @@ def setup_postgres_relation(harness):
...
+        "postgresql-k8s",
+        {
+            "data": '{"database": "database", "extra-user-roles": "SUPERUSER"}',
+            "endpoints": DB_ENDPOINTS,
+        },
+    )
+
+
+def test_correct_config(harness, mocked_kubernetes_service_patcher, mocked_sql_migration) -> None:
     harness.begin()
```
gruyaume, line 42: You can place the `begin` in a setUp it's probably used in each test

tests/unit/test_charm.py
```diff
@@ -85,59 +97,124 @@ def test_update_container_correct_pebble_layer(
...
+        }
+    }
+    updated_plan = harness.get_container_pebble_plan(CONTAINER_NAME).to_dict()
+    assert expected_plan == updated_plan
+    updated_config = yaml.safe_load(harness.charm._config)
+    assert DB_ENDPOINTS in updated_config["dsn"]
+    assert DB_PASSWORD in updated_config["dsn"]
+    assert DB_USERNAME in updated_config["dsn"]
+
+    service = harness.model.unit.get_container("kratos").get_service("kratos")
```
gruyaume, line 136: Don't validate too many things in 1 test. A unit test should only have 1 reason to fail.

tests/unit/test_charm.py
```diff
@@ -85,59 +97,124 @@ def test_update_container_correct_pebble_layer(
...
 
-    mocked_update_container.assert_called_once()
+    trigger_database_changed(harness)
+
+    assert isinstance(harness.charm.unit.status, WaitingStatus)
 
 
-def test_on_database_created(
-    harness, mocked_kubernetes_service_patcher, mocked_update_container
+def test_on_database_changed_after_pebble_ready(
```
gruyaume, line 207: I would recommend using `given_x_when_y_then_z` as a test naming convention. This avoids using words like `with` and `after` and makes it obvious for readers of the unit tests what the expected behaviour is

src/charm.py
```diff
@@ -176,13 +122,71 @@ def _run_sql_migration(self) -> None:
...
+        with open("src/identity.default.schema.json", encoding="utf-8") as schema_file:
+            schema = schema_file.read()
+            self._container.push(self._identity_schema_file_path, schema, make_dirs=True)
+        self._container.add_layer(self._container_name, self._pebble_layer, combine=True)
+        logger.info("Pebble plan updated with new configuration, replanning")
+        self._container.replan()
+        if self.database.is_database_created():
+            self._container.push(self._config_file_path, self._config, make_dirs=True)
+            self._container.start(self._container_name)
+        self.unit.status = ActiveStatus()
```
natalian98, line 142: The status will be set to active even if database details can not retrieved and thus container isn't started

tests/unit/test_charm.py
agathanatasha: The next test asserts the empty initial pebble plan before beginning

src/charm.py
```diff
@@ -97,92 +97,113 @@ def _config(self) -> str:
...
+        self._container.add_layer(self._container_name, self._pebble_layer, combine=True)
+        logger.info("Pebble plan updated with new configuration, replanning")
+        self._container.replan()
+
+        # in case container was terminated unexpectedly
+        if self.database.is_database_created():
+            self._container.push(self._config_file_path, self._config, make_dirs=True)
+            self._container.start(self._container_name)
+            self.unit.status = ActiveStatus()
             return
```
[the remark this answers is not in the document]
agathanatasha: I think toggling layer settings like this is an anti-pattern. I am okay with pushing the config file regardless.

src/charm.py
```diff
@@ -138,13 +144,13 @@ def _on_pebble_ready(self, event) -> None:
         with open("src/identity.default.schema.json", encoding="utf-8") as schema_file:
             schema = schema_file.read()
             self._container.push(self._identity_schema_file_path, schema, make_dirs=True)
+        self._container.push(self._config_file_path, self._config, make_dirs=True)
```
natalian98, line 147: The config needs dsn with database details to be able to start the service. Does it make sense to push it when database is not yet created?

src/charm.py
```diff
@@ -97,92 +100,116 @@ def _config(self) -> str:
...
-                self.unit.status = MaintenanceStatus("Updating Kratos Config")
-                self._container.push(self._config_file_path, self._config, make_dirs=True)
-                logger.info("Updated kratos config")
-                self._container.restart(self._container_name)
-
     def _get_database_relation_info(self) -> dict:
         """Get database info from relation data bag."""
         relation_id = self.database.relations[0].id
+        print(
+            f"DEBUGGING ~ file: charm.py ~ line 103 ~ self.database.relations: {self.database.relations}"
```
natalian98, line 107: Is it still relevant? Should there be a log entry instead?

.github/workflows/on_pull_request.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Install dependencies
-        run: python3 -m pip install tox
+        run: python3 -m pip install tox==3.27.1
```
natalian98, line 20: I think the tox version doesn't need to be pinned if you fix [this](https://github.com/canonical/kratos-operator/pull/22/files#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449R39) in `tox.ini` file

src/charm.py
```diff
@@ -97,92 +100,116 @@ def _config(self) -> str:
...
             return
 
-        if not self.unit.is_leader():
-            return
+        logger.info("Updating Kratos config and restarting service")
+        self._container.push(self._config_file_path, self._config, make_dirs=True)
 
-        self._run_sql_migration()
+        if self.unit.is_leader() and not self._run_sql_migration():
+            self.unit.status = BlockedStatus("Database migration failed.")
```
natalian98, line 187: Would you like to add an action for running the migration or handle it in another PR?

natalian98, line 187: Would you like to add an action for running the migration or handle it in another PR?

agathanatasha: in another pr

src/charm.py
agathanatasha: good catch, thanks

src/charm.py
agathanatasha: It was a compromise made earlier with Nikos' suggestion. I prefer to only push if database is created. I think this is also an okay approach.

nsklikas: It was just a suggestion to make the flow simpler. If you prefer the previous approach I have no objections.

natalian98: It's ok, I also ended up using this approach in hydra.

src/charm.py
```diff
@@ -97,92 +100,113 @@ def _config(self) -> str:
...
 
-        if not self.model.relations["pg-database"]:
-            logger.error("Missing required relation with postgresql")
-            self.model.unit.status = BlockedStatus("Missing required relation with postgresql")
+        self.unit.status = MaintenanceStatus("Configuring/deploying resources")
+
+        with open("src/identity.default.schema.json", encoding="utf-8") as schema_file:
+            schema = schema_file.read()
+            self._container.push(self._identity_schema_file_path, schema, make_dirs=True)
+        self._container.push(self._config_file_path, self._config, make_dirs=True)
```
gruyaume, line 144: Why do we push this file on pebble ready and not only on database created/changed event?

src/charm.py
```diff
@@ -97,92 +100,113 @@ def _config(self) -> str:
...
 
-        self._run_sql_migration()
+        if self.unit.is_leader() and not self._run_sql_migration():
+            self.unit.status = BlockedStatus("Database migration failed.")
+        else:
+            self._container.start(self._container_name)
+            self.unit.status = ActiveStatus()
 
-        self.unit.status = ActiveStatus()
+    def _on_database_changed(self, event: DatabaseCreatedEvent) -> None:
```
gruyaume, line 189:
```suggestion
    def _on_database_changed(self, event: DatabaseEndpointsChangedEvent) -> None:
```

agathanatasha:
replace by #24

PR #20, opened 2022-12-06
Add external provider support

nsklikas:
Use the external-provider integrator

TODO:
- [x] Add the rest of the claim mappers


natalian98:
Overall I like it, good job. Since you are providing a new relation, maybe you could add an integration test for it as well?
I'd also update the usage part in readme and at least link the external integrator.
If I understand correctly, there will be multiple claim mappers, one per each IdP? How will they be added?

src/charm.py
```diff
@@ -102,22 +115,42 @@ def _pebble_layer(self) -> Layer:
...
 
+    @property
+    def _domain_url(self):
+        return self.config["external_url"] or self.public_ingress.url
+
+    @cached_property
+    def _get_available_mappers(self):
+        return [
+            schema_file.name[: -len("_schema.jsonnet")]
+            for schema_file in Path("claim_mappers").iterdir()
```
natalian98, line 126: You could make it a variable since you use this path multiple times

nsklikas:
> Since you are providing a new relation, maybe you could add an integration test for it as well? 

I didn't add an integration test, because IMO we would have to test that the flow works for it to make sense, which means that we would have to actually login to an external provider (and this requires UI). The unit tests should be enough for now.

> If I understand correctly, there will be multiple claim mappers, one per each IdP? How will they be added?

There is a default claim mapper that will be used, if we want a special mapper for a specific provider we will have to add it to this repo.

natalian98:
> > Since you are providing a new relation, maybe you could add an integration test for it as well?
> 
> I didn't add an integration test, because IMO we would have to test that the flow works for it to make sense, which means that we would have to actually login to an external provider (and this requires UI). The unit tests should be enough for now.

Fair point, unit tests will be enough for now. Just reference the relation in readme.

natalian98 (approved):
lgtm

PR #22, opened 2022-12-14
Add ingress

nsklikas:
Add ingress relation.

To test this on a model that has kratos deployed you need to run (we need `channel=latest/edge` due to https://github.com/canonical/traefik-k8s-operator/pull/114):
```console
juju deploy traefik-k8s traefik-public --channel latest/edge
juju deploy traefik-k8s traefik-admin --channel latest/edge
juju config traefik-public external_hostname=some_host
juju config traefik-admin external_hostname=some_host
juju relate traefik-public kratos:public-ingress
juju relate traefik-admin kratos:admin-ingress
```

After all models in the unit reach an active state, get the traefik IPs by running:
```console
juju status
...
traefik-admin/0*                  active       idle   {admin_IP}         
traefik-public/0*                  active       idle   {public_IP}         
```
Then on your browser you should be able to reach Kratos APIs through these IPs (eg `http://{admin_IP}/{juju_model_name}-kratos/admin/identities` and `http://{public_IP}/{juju_model_name}-kratos/.well-known/ory/webauthn.js`). 

Currently there is a problem with the traefik charm and all traffic will be redirected to the Kratos leader.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -22,13 +22,12 @@ from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingSta
 from ops.pebble import ChangeError, ExecError, Layer
 
 logger = logging.getLogger(__name__)
+KRATOS_ADMIN_PORT = 4434
+KRATOS_PUBLIC_PORT = 4433
 
 
 class KratosCharm(CharmBase):
     """Charmed Ory Kratos."""
-
-    admin_port = 4434
-    public_port = 4433
 
     def __init__(self, *args):
         super().__init__(*args)
@@ -40,18 +39,18 @@ class KratosCharm(CharmBase):
         self._db_name = f"{self.model.name}_{self.app.name}"
 
         self.service_patcher = KubernetesServicePatch(
-            self, [("admin", self.admin_port), ("public", self.public_port)]
+            self, [("admin", KRATOS_ADMIN_PORT), ("public", KRATOS_PUBLIC_PORT)]
         )
         self.admin_ingress = IngressPerAppRequirer(
             self,
             relation_name="admin-ingress",
-            port=self.admin_port,
+            port=KRATOS_ADMIN_PORT,
             strip_prefix=True,
         )
         self.public_ingress = IngressPerAppRequirer(
             self,
             relation_name="public-ingress",
-            port=self.public_port,
+            port=KRATOS_PUBLIC_PORT,
             strip_prefix=True,
         )
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -66,11 +66,11 @@ class KratosCharm(CharmBase):
         self.framework.observe(self.database.on.endpoints_changed, self._on_database_changed)
 
         # Admin ingress events
-        self.framework.observe(self.admin_ingress.on.ready, self._on_ingress_ready)
+        self.framework.observe(self.admin_ingress.on.ready, self._on_admin_ingress_ready)
         self.framework.observe(self.admin_ingress.on.revoked, self._on_ingress_revoked)
 
         # Public ingress events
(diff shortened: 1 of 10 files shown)


src/charm.py
```diff
@@ -22,14 +27,39 @@
 class KratosCharm(CharmBase):
     """Charmed Ory Kratos."""
 
+    admin_port = 4434
```
natalian98, line 30: Could we make these port variables uppercase on module level as they are constants?

src/charm.py
```diff
@@ -22,14 +27,39 @@
...
+            strip_prefix=True,
+        )
+        self.public_ingress = IngressPerAppRequirer(
+            self,
+            relation_name="public-ingress",
+            port=self.public_port,
+            strip_prefix=True,
+        )
+
+        # Admin ingress events
```
natalian98, line 56: I find it more readable to put all observed events together rather than splitting in ingress, db and pebble-related, but it's just a minor suggestion and you can dismiss it. Let's see what others think

src/charm.py
```diff
@@ -184,6 +214,15 @@ def _on_database_changed(self, event: DatabaseCreatedEvent) -> None:
         self.unit.status = MaintenanceStatus("Retrieving database details")
         self._update_container(event)
 
+    def _on_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
+        if self.unit.is_leader():
```
natalian98, line 218: > Currently there is a problem with the traefik charm and all traffic will be redirected to the Kratos leader.

Will it all be handled on traefik side? no action needed on kratos?

natalian98:
Would it be possible to add some integration test for this relation and/or update the readme?
I also see that the traefik fix was merged (rev98), maybe we could add a note about minimal revision?

src/charm.py
nsklikas: Yes, I think that no changes (or very little) will be needed on the Kratos side to fix this.

src/charm.py
```diff
@@ -42,6 +63,14 @@ def __init__(self, *args):
         self.framework.observe(self.database.on.database_created, self._on_database_changed)
         self.framework.observe(self.database.on.endpoints_changed, self._on_database_changed)
 
+        # Admin ingress events
```
gruyaume, line 66: I don't think those comments add to the understanding

src/charm.py
```diff
@@ -184,6 +213,15 @@ def _on_database_changed(self, event: DatabaseCreatedEvent) -> None:
         self.unit.status = MaintenanceStatus("Retrieving database details")
         self._update_container(event)
 
+    def _on_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
+        if self.unit.is_leader():
+            type = "admin" if event.relation_name == self.admin_ingress.relation_name else "public"
```
gruyaume, line 218: `type` is not a great variable name since it is also a python built-in function. Could use `ingress_type`.

gruyaume, line 218: Also, if handling admin and public ingress is not exactly the same, why not have two different event methods instead of combining them into `_on_ingress_ready`. More code is not a bad thing, what we want to reduce is complexity.

src/charm.py
nsklikas: For me they are useful when I am trying to skim through the registered hooks of an operator that has many of them. If you wish I can remove these comments.

agathanatasha (dismissed):

src/charm.py
gruyaume: I don't insist on it, honestly it doesn't make much of a difference, but if we want to have comments to separate the event types, we should probably add some to the database events as well to at least be consistent.

gruyaume (approved):

natalian98 (dismissed):

natalian98 (approved):

PR #25, opened 2023-01-05
Database migrated peer relation flag

agathanatasha:


nsklikas:
I don't really like the `PeerRelationState` class that is used in the openfga charm, I would prefer if didn't use it.

I am not sure about when/if the peer relation databag can be cleared, but IMO we should make `STATE_KEY_IS_DB_MIGRATED` an `str` rather than `bool` and store the version of the migration there (we can use the oci-image id if the ory cli does not provide something), so that we can also handle image upgrades this way.

nsklikas (changes requested):

PR #28, opened 2023-01-18
Add dev flag

nsklikas:
Adds configuration option for using the `dev` flag when starting Kratos. This flag is needed when running Kratos without HTTPS

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -11,6 +11,6 @@ options:
     default: ""
   dev:
     description:
-      Run Kratos on dev mode, it is needed if HTTPS is not set up. This should only be used for developement purposes.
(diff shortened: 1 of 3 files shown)


config.yaml
```diff
@@ -9,3 +9,8 @@ options:
       if Kratos is served via a reverse proxy).
     type: string
     default: ""
+  dev:
+    description:
+      Run Kratos on dev mode, it is needed if HTTPS is not set up. This should only be used for developement purposes.
```
natalian98, line 14: that should solve the lint error
```suggestion
      Run Kratos on dev mode, it is needed if HTTPS is not set up. This should only be used for development purposes.
```

natalian98 (approved):

PR #32, opened 2023-02-08
Initial Hydra and UI integration

natalian98:
Initial integration with hydra and login ui charms:
- bumped kratos image to allow for the integration
- added hydra-kratos relation interface to get `oauth2_provider_url` (hydra admin API endpoint)
- updated kratos config
- updated unit tests
- updated readme according to the current state of integration

Integration testing will be covered in bundle tests.

### DIFF
--- a/lib/charms/hydra/v0/hydra_endpoints.py
+++ b/lib/charms/hydra/v0/hydra_endpoints.py
@@ -46,9 +46,7 @@ import logging
 
 from ops.charm import (
     CharmBase,
-    RelationChangedEvent,
     RelationCreatedEvent,
-    RelationJoinedEvent,
 )
 from ops.framework import EventBase, EventSource, Object, ObjectEvents
 from ops.model import Application
@@ -90,23 +88,13 @@ class HydraEndpointsProvider(Object):
         self._relation_name = relation_name
 
         events = self._charm.on[relation_name]
-        self.framework.observe(events.relation_joined, self._on_provider_endpoint_relation_joined)
-        self.framework.observe(events.relation_created, self._on_provider_endpoint_relation_joined)
-        self.framework.observe(
-            events.relation_changed, self._on_provider_endpoint_relation_changed
-        )
-
-    def _on_provider_endpoint_relation_joined(self, event: RelationJoinedEvent):
-        self.on.ready.emit()
+        self.framework.observe(events.relation_created, self._on_provider_endpoint_relation_created)
 
     def _on_provider_endpoint_relation_created(self, event: RelationCreatedEvent):
         self.on.ready.emit()
 
-    def _on_provider_endpoint_relation_changed(self, event: RelationChangedEvent):
-        self.on.ready.emit()
-
     def send_endpoint_relation_data(
-        self, charm, admin_endpoint: str, public_endpoint: str
+        self, charm: CharmBase, admin_endpoint: str, public_endpoint: str
     ) -> None:
         """Updates relation with endpoints info."""
         if not self._charm.unit.is_leader():
--- a/lib/charms/hydra/v0/hydra_endpoints.py
+++ b/lib/charms/hydra/v0/hydra_endpoints.py
@@ -68,14 +68,14 @@ INTERFACE_NAME = "hydra_endpoints"
 logger = logging.getLogger(__name__)
 
 
-class RelationReadyEvent(EventBase):
+class HydraEndpointsRelationReadyEvent(EventBase):
     """Event to notify the charm that the relation is ready."""
 
 
 class HydraEndpointsProviderEvents(ObjectEvents):
     """Event descriptor for events raised by `HydraEndpointsProvider`."""
 
-    ready = EventSource(RelationReadyEvent)
+    ready = EventSource(HydraEndpointsRelationReadyEvent)
 
 
 class HydraEndpointsProvider(Object):
--- a/lib/charms/hydra/v0/hydra_endpoints.py
+++ b/lib/charms/hydra/v0/hydra_endpoints.py
@@ -10,7 +10,7 @@ This library provides a Python API for both requesting and providing public and 
 To get started using the library, you need to fetch the library using `charmcraft`.
 ```shell
 cd some-charm
-charmcraft fetch-lib charms.hydra.v0.hydra_endpoints_info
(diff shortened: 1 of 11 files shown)


src/charm.py
```diff
@@ -162,6 +177,18 @@ def _push_schemas(self):
                 schema = f.read()
             self._container.push(Path(self._config_dir_path, schema_file), schema, make_dirs=True)
 
+    def _get_hydra_endpoint_info(self) -> str:
+        # Assign a default value as config file won't accept an empty string
+        oauth2_provider_url = "http://127.0.0.1:4445/"
```
nsklikas, line 182: This default does not make sense in the context of k8s. I think it would be better to just return an `None` oauth2_provider_url and handle it in the config template.

If `oauth2_provider_url` is not set, then the template should not add the `oauth2_provider_url` the config file, e.g.:
```
{%- if oauth2_provider_url %}
oauth2_provider:
    url: {{ oauth2_provider_url }}
{%- endif %}
```

lib/charms/hydra/v0/hydra_endpoints.py
```diff
@@ -0,0 +1,172 @@
...
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+RELATION_NAME = "endpoint-info"
+INTERFACE_NAME = "hydra_endpoints"
+logger = logging.getLogger(__name__)
+
+
+class RelationReadyEvent(EventBase):
```
gruyaume, line 71: `RelationReadyEvent` is not specific enough imo, I'd replace it with `HydraEndpointsRelationReadyEvent` . Here the charm will have to use this relation and if Hydra provides multiple relation interfaces, `RelationReadyEvent` wouldn't tell us which relation is ready.

src/charm.py
```diff
@@ -162,6 +177,17 @@ def _push_schemas(self):
                 schema = f.read()
             self._container.push(Path(self._config_dir_path, schema_file), schema, make_dirs=True)
 
+    def _get_hydra_endpoint_info(self) -> None | str:
```
gruyaume, line 180: Typically, we use `Optional[str]` instead of `None | str` for the return typing

src/charm.py
```diff
@@ -162,6 +177,17 @@ def _push_schemas(self):
...
+        oauth2_provider_url = None
+        if self.model.relations[self._hydra_relation_name]:
+            try:
+                hydra_endpoints = self.hydra_endpoints.get_hydra_endpoints()
+                oauth2_provider_url = hydra_endpoints["admin_endpoint"]
+            except HydraEndpointsRelationDataMissingError:
+                logger.info("No hydra endpoint-info relation data found")
```
gruyaume, line 187: add `return None` to make it obvious we don't return anything

src/charm.py
```diff
@@ -146,12 +158,15 @@ def _render_conf_file(self) -> None:
...
+            default_browser_return_url=self.config.get("login_ui_url"),
+            public_base_url=self._domain_url,
+            login_ui_url=join(self.config.get("login_ui_url"), "login"),
+            error_ui_url=join(self.config.get("login_ui_url"), "oidc_error"),
             oidc_providers=self.external_provider.get_providers(),
             available_mappers=self._get_available_mappers,
-            registration_ui_url="http://127.0.0.1:9999/registration",
+            registration_ui_url=join(self.config.get("login_ui_url"), "registration"),
             db_info=self._get_database_relation_info(),
+            oauth2_provider_url=self._get_hydra_endpoint_info(),
```
gruyaume, line 169: Here we sould simetimes render a template with an empty `oauth2_provider_url`, is this normal?

tests/unit/test_charm.py
```diff
@@ -51,6 +51,20 @@ def setup_peer_relation(harness):
     )
 
 
+def setup_hydra_relation(harness):
```
gruyaume, line 54: Add return typing

tests/unit/test_charm.py
```diff
@@ -516,3 +569,103 @@ def test_on_client_config_changed_with_external_url_config(harness, mocked_conta
...
+        },
+        "oauth2_provider": {
+            "url": "http://hydra-admin-url:80/testing-hydra",
+        },
+    }
+
+    assert yaml.safe_load(harness.charm._render_conf_file()) == expected_config
+
+
+def test_on_client_config_changed_when_missing_hydra_relation_data(harness) -> None:
```
gruyaume, line 625: Pretty close to a `given_x_when_y_then_z`!! I'd definitely encourage going that route :) Here it'd look like

`given_missing_hydra_relation_data_when_config_changed_then_config_file_is_rendered`

tests/unit/test_charm.py
```diff
@@ -516,3 +569,103 @@ def test_on_client_config_changed_with_external_url_config(harness, mocked_conta
...
+            "public": {
+                "base_url": "None",
+                "cors": {
+                    "enabled": True,
+                },
+            },
+        },
+    }
+
+    assert yaml.safe_load(harness.charm._render_conf_file()) == expected_config
```
gruyaume, line 671: Why test for the output of this call? We should probably test for the fact that a file with this content was pushed on the workload? If we make a change to that method like change its name or something then the test would crash

src/charm.py
natalian98: If `oauth2_provider_url` is empty, jinja template will not add it to the config:
```
{%- if oauth2_provider_url %}
oauth2_provider:
    url: {{ oauth2_provider_url }}
{%- endif %}
```

tests/unit/test_charm.py
natalian98: I was trying to follow the existing naming convention, should I rename all tests?

tests/unit/test_charm.py
natalian98: Done, thanks for the suggestion

tests/unit/test_charm.py
gruyaume: Ultimately I think this naming convention is a good one (`given_x_when_y_then_z`) but it's not as if it was widely used at Canonical so I don't wan't to force it on you.

nsklikas (dismissed):

gruyaume (approved):

nsklikas (approved):

PR #48, opened 2023-04-03
Add actions

nsklikas:
- [x] The `reset-password` and `create-admin-account` (without a password provided) will not work until we implement the settings and recovery endpoints, should we just comment the actions out until that happens?
- [x] The `run-migration` action is not sufficiently tested, perhaps we should open an issue to look into that in the future

### DIFF
--- a/actions.yaml
+++ b/actions.yaml
@@ -54,7 +54,7 @@ create-admin-account:
       description: The admin's phone number
       type: string
   # TODO: Remove required password once password resetting is implemented
-  required: [username, password]
+  required: ["username", "password"]
 run-migration:
   description: |
     Run a migration, this is needed after upgrades. This is a non-reversible operation.
--- a/actions.yaml
+++ b/actions.yaml
@@ -16,21 +16,6 @@ delete-identity:
     email:
       description: The user's email
       type: string
-# reset-password:
-#   description: |
-#     Reset a user’s using either the identity ID or the user email, either a magic
-#     link or link and a one time code will be returned where the user will be asked to
-#     provide a new password
-#   params:
-#     identity-id:
-#       description: The Identity ID
-#       type: string
-#     email:
-#       description: The user's email
-#       type: string
-#     recovery-method:
-#       description: Whether a magic link or a one-time code will be used for resetting
-#       default: "link"
 create-admin-account:
   description: |
     Create an admin user. If no password was provided, the command will return a
--- a/actions.yaml
+++ b/actions.yaml
@@ -16,21 +16,21 @@ delete-identity:
     email:
       description: The user's email
       type: string
-reset-password:
-  description: |
-    Reset a user’s using either the identity ID or the user email, either a magic
-    link or link and a one time code will be returned where the user will be asked to
-    provide a new password
-  params:
-    identity-id:
-      description: The Identity ID
-      type: string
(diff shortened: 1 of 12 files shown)


actions.yaml
```diff
@@ -0,0 +1,66 @@
...
+      description: The admin's password
+      type: string
+    name:
+      description: The admin's name
+      type: string
+    phone_number:
+      description: The admin's phone number
+      type: string
+  # TODO: Remove required password once password resetting is implemented
+  required: [username, password]
```
natalian98, line 57:
```suggestion
  required: ["username", "password"]
```

natalian98 (commented):
Could you list the actions in readme?

src/charm.py
```diff
@@ -424,6 +470,154 @@ def _on_client_config_changed(self, event: ClientConfigChangedEvent) -> None:
...
+        event.log("Follow the link to reset the user's password")
+        event.set_results(dict_to_action_output(ret))
+
+    def _on_create_admin_account_action(self, event: ActionEvent) -> None:
+        if not self._kratos_service_is_running:
+            event.fail("Service is not ready. Please re-run the action when the charm is active")
+            return
+
+        traits = {
+            "username": event.params["username"],
```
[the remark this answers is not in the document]
nsklikas: `username` is a mandatory parameter, we don't need to use `get` because it will always be there.

actions.yaml
```diff
@@ -0,0 +1,51 @@
...
+      type: string
+    name:
+      description: The admin's name
+      type: string
+    phone_number:
+      description: The admin's phone number
+      type: string
+  # TODO: Remove required password once password resetting is implemented
+  required: ["username", "password"]
+run-migration:
```
jnsgruk, line 43: Should this not be done automatically as part of the upgrade process?

nsklikas: I think this was the data platform team's suggestion for handling migrations. 

It may be a breaking migration, so the admin will have to first take a backup. In addition, if the charm tries to run a migration and it fails for some reason, the admin would need to be able to trigger it again (e.g. with bigger timeout or after some transient issue has been resolved).

jnsgruk: This is *sort of the point* of operators, though, precisely to automate this kind of stuff

Sure, the upgrade guide might say:

- Run your DB backup
- Refresh the Kratos charm

But I still think this should be handled automatically

tests/unit/test_kratos.py
```diff
@@ -0,0 +1,218 @@
+# Copyright 2022 Canonical Ltd.
```
natalian98, line 1:
```suggestion
# Copyright 2023 Canonical Ltd.
```

natalian98 (commented):
Looks good to me, I really like the tests. Only some minor comments.

Regarding the action for migration, I think we should create a separate task for it and discuss with data platform team again if we can trigger the backup before migration.

actions.yaml
nsklikas: I don't disagree with this, but we chose this approach as per the data platform team's suggestion (as I mentioned before). 

Having said that, it is out of scope for this PR to handle the db migration logic. We will create a task in our backlog for investigating this in the future.

natalian98 (dismissed):

PR #49, opened 2023-04-04
feat: requirer side implementation of login_ui_endpoints interface 

bencekov:
- implemented requirer side of login_ui_endpoints
- edited login_ui_url out of config.yaml
- updated unit tests
- updated README.md


README.md
```diff
@@ -94,17 +94,14 @@ juju integrate kratos hydra
...
 
 ### Identity Platform Login UI
-<!-- TODO: Change this section when identity-platform-login-ui-operator endpoints relation is ready -->
 
-If you have deployed [Login UI charm](https://github.com/canonical/identity-platform-login-ui-operator), you can configure it with kratos by providing its URL.
-Note that the UI charm should run behind a proxy.
-```console
-juju config kratos kratos_ui_url=http://{traefik_public_ip}/{model_name}-{identity_platform_login_ui_app_name}
-```
+The following instructions assume that you have deployed `traefik-admin` and `traefik-public` charms and related them to hydra. Note that the UI charm should run behind a proxy.
```
natalian98, line 98:
```suggestion
The following instructions assume that you have deployed `traefik-admin` and `traefik-public` charms and related them to kratos. Note that the UI charm should run behind a proxy.
```

src/charm.py
```diff
@@ -165,13 +177,13 @@ def _render_conf_file(self) -> str:
         rendered = template.render(
             mappers_path=str(self._mappers_dir_path),
             identity_schema_file_path=self._identity_schema_file_path,
-            default_browser_return_url=self.config.get("login_ui_url"),
+            default_browser_return_url=self._get_login_ui_endpoint_info("default_url"),
             public_base_url=self._domain_url,
-            login_ui_url=join(self.config.get("login_ui_url"), "login"),
-            error_ui_url=join(self.config.get("login_ui_url"), "oidc_error"),
+            login_ui_url=self._get_login_ui_endpoint_info("login_url"),
+            error_ui_url=self._get_login_ui_endpoint_info("error_url"),
```
natalian98, line 183: Kratos won't accept an empty value for a defined config option. So if any of the endpoints wasn't found and thus was returned as `None`, kratos would error out due to invalid config.
You could try overcoming it by modifying the template and not adding the config if the endpoint is `None`, like we do [here](https://github.com/canonical/kratos-operator/blob/main/templates/kratos.yaml.j2#L54) with the `oauth2_provider_url`.
However you need to test if kratos will start its service without any of the login-ui's endpoints.

README.md
bencekov: done

src/charm.py
bencekov: I updated the template

templates/kratos.yaml.j2
```diff
@@ -5,14 +5,23 @@ identity:
     schemas:
         - id: default
           url: 'file://{{ identity_schema_file_path }}'
+{%- if default_browser_return_url or oidc_providers %}
 selfservice:
+    {%- if default_browser_return_url %}
```
natalian98, line 10: This parameter is required to start kratos, so we should always add it. I think that's why the tests are failing. Besides, `oidc_providers` has no correlation in this section
```suggestion
selfservice:
```

natalian98: To recap our conversations on mm or dailies around that topic, we should have a default url set in kratos for `default_browser_return_url`, no matter if the `login-endpoint-info` relation or its data is in place or not.

For other parameters that are not required to start kratos and run basic commands like creating identities, we can keep it as you've done in the template.

bencekov: done

src/charm.py
```diff
@@ -198,6 +212,26 @@ def _get_hydra_endpoint_info(self) -> Optional[str]:
...
 
+    def _get_login_ui_endpoint_info(self, key: str) -> Optional[str]:
+        if self.model.relations[self._login_ui_relation_name]:
+            try:
+                login_ui_endpoints = self.login_ui_endpoints.get_login_ui_endpoints()
+                login_ui_url = login_ui_endpoints[key]
+                return login_ui_url
+            except LoginUIEndpointsRelationDataMissingError:
+                logger.info("No login ui endpoint-info relation data found")
+                if key == "default_url":
```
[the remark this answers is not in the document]
bencekov: Well, we want to use the constant DEFAULT_BROWSER_RETURN_URL, if there's no relation between kratos-operator and an instance of login-ui operator, or there's a problem with accessing login-ui endpoints (for example there's no public facing proxy). I get that it looks bloated the way it is right now, but checking these conditions are necessary, because we only want to use the default if something goes wrong. Taking key=="default_url" to the root of the method would then replicate the nested condition checking.

With that being said, I can change it, but returning the default every time the key is "default_url" is not an option.

tests/unit/test_charm.py
```diff
@@ -646,17 +704,25 @@ def test_on_client_config_changed_when_missing_hydra_relation_data(harness) -> N
...
         },
         "selfservice": {
-            "default_browser_return_url": "http://127.0.0.1:4455/",
+            "default_browser_return_url": harness.get_relation_data(
+                login_relation_id, "identity-platform-login-ui-operator"
+            )["default_url"],
             "flows": {
                 "error": {
-                    "ui_url": "http://127.0.0.1:4455/oidc_error",
+                    "ui_url": harness.get_relation_data(
```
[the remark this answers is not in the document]
bencekov: I changed the tests to reflect your request.

src/charm.py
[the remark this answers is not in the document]
nsklikas: I would suggest to handle this in the jinja template with something like:
```
         {{ default_browser_return_url | d("http://example-default-return-url.com") }}
```

src/charm.py
```diff
@@ -198,6 +211,21 @@ def _get_hydra_endpoint_info(self) -> Optional[str]:
 
         return oauth2_provider_url
 
+    def _get_login_ui_endpoint_info(self, key: str) -> Optional[str]:
+        if self.model.relations[self._login_ui_relation_name]:
```
nsklikas, line 215: The login_ui_endpoints library handles this, no need to check it here.

src/charm.py
```diff
@@ -165,13 +178,13 @@ def _render_conf_file(self) -> str:
         rendered = template.render(
             mappers_path=str(self._mappers_dir_path),
             identity_schema_file_path=self._identity_schema_file_path,
-            default_browser_return_url=self.config.get("login_ui_url"),
+            default_browser_return_url=self._get_login_ui_endpoint_info("default_url"),
             public_base_url=self._domain_url,
-            login_ui_url=join(self.config.get("login_ui_url"), "login"),
-            error_ui_url=join(self.config.get("login_ui_url"), "oidc_error"),
+            login_ui_url=self._get_login_ui_endpoint_info("login_url"),
```
nsklikas, line 183: Just a suggestion: 
`self._get_login_ui_endpoint_info` could return an object that has all the endpoints. And then extract the necessary info in the jinja template. 

This way we could be consistent in how we handle data from relation interfaces (that's how we do it for `_get_database_relation_info` and `_get_hydra_endpoint_info`)

src/charm.py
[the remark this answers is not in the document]
bencekov: changed charm.py and jinja template. default browser is now handled by the template. I used an if else statement, because an logical or printed "None" value.

src/charm.py
bencekov: done. (included LoginUIEndpointsRelationMissingError handling)

lib/charms/identity_platform_login_ui_operator/v0/login_ui_endpoints.py
```diff
@@ -0,0 +1,165 @@
...
+            events.relation_created, self._on_provider_endpoints_relation_created
+        )
+
+    def _on_provider_endpoints_relation_created(self, event: RelationCreatedEvent) -> None:
+        self.on.ready.emit()
+
+    def send_endpoints_relation_data(self, endpoint: str) -> None:
+        """Updates relation with endpoint info."""
+        if not self._charm.unit.is_leader():
+            return
```
gruyaume, line 103: Since this is in a library, we may want to raise here (instead of returning) so that the user knows has feedback that they shouldn't call this method from non-leader units. And in the charm you will want to make sure that it's only the leader that would call this method.

lib/charms/identity_platform_login_ui_operator/v0/login_ui_endpoints.py
```diff
@@ -0,0 +1,165 @@
...
+    """Requirer side of the ui-endpoint-info relation."""
+
+    def __init__(self, charm: CharmBase, relation_name: str = RELATION_NAME):
+        super().__init__(charm, relation_name)
+        self.charm = charm
+        self._relation_name = relation_name
+
+    def get_login_ui_endpoints(self) -> Optional[Dict]:
+        """Get the Identity Platform Login UI endpoints."""
+        endpoints = self.model.relations[self._relation_name]
```
gruyaume, line 152: Not a great variable name, `ui_endpoint_relations` would make more sense

lib/charms/identity_platform_login_ui_operator/v0/login_ui_endpoints.py
```diff
@@ -0,0 +1,165 @@
...
+        if len(endpoints) == 0:
+            raise LoginUIEndpointsRelationMissingError()
+
+        if not (app := endpoints[0].app):
+            raise LoginUIEndpointsRelationMissingError()
+        data = endpoints[0].data[endpoints[0].app]
```
gruyaume, line 158: Again, not great variable name, `ui_endpoint_relation_data` would be better

lib/charms/identity_platform_login_ui_operator/v0/login_ui_endpoints.py
bencekov: The function now raises an error, which is handled in charm.py

lib/charms/identity_platform_login_ui_operator/v0/login_ui_endpoints.py
bencekov: done

lib/charms/identity_platform_login_ui_operator/v0/login_ui_endpoints.py
bencekov: done

lib/charms/identity_platform_login_ui_operator/v0/login_ui_endpoints.py
```diff
@@ -0,0 +1,165 @@
...
+        self.charm = charm
+        self._relation_name = relation_name
+
+    def get_login_ui_endpoints(self) -> Optional[Dict]:
+        """Get the Identity Platform Login UI endpoints."""
+        endpoints = self.model.relations[self._relation_name]
+        if len(endpoints) == 0:
+            raise LoginUIEndpointsRelationMissingError()
+
+        if not (app := endpoints[0].app):
```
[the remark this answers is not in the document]
bencekov: I changed the method to check the app for all relations in ui_endpoint_relations

nsklikas (commented):
LGTM, other than a couple of comments

tests/unit/test_charm.py
```diff
@@ -69,6 +70,26 @@ def setup_hydra_relation(harness: Harness) -> int:
     return relation_id
 
 
+def setup_login_ui_relation(harness) -> int:
```
natalian98, line 73:
```suggestion
def setup_login_ui_relation(harness: Harness) -> int:
```

tests/unit/test_charm.py
```diff
@@ -743,6 +836,93 @@ def test_kratos_endpoint_info_relation_data_with_ingress_relation_data(harness:
...
+        "oauth2_provider": {
+            "url": "http://hydra-admin-url:80/testing-hydra",
+        },
+    }
+
+    container_config = container.pull(path="/etc/config/kratos.yaml", encoding="utf-8")
+    assert yaml.load(container_config.read(), yaml.Loader) == expected_config
+
+
+def test_on_client_config_changed_when_missing_login_ui_and_hydra_relation_data(harness) -> None:
```
natalian98, line 883:
```suggestion
def test_on_client_config_changed_when_missing_login_ui_and_hydra_relation_data(harness: Harness) -> None:
```

natalian98 (commented):
lgtm, let's just add the missing typing in unit tests
Also, please use squash and merge option while merging this PR after it gets the approval.

lib/charms/identity_platform_login_ui_operator/v0/login_ui_endpoints.py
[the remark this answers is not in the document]
bencekov: The method now uses get_relation, and raises an error if there are more than one relations.

templates/kratos.yaml.j2
```diff
@@ -9,12 +9,22 @@ identity:
           url: 'file://{{ admin_identity_schema_file_path }}'
 selfservice:
     default_browser_return_url:
+        {%- if default_browser_return_url %}
         {{ default_browser_return_url }}
+        {% else %}
+        http://example-default-return-url.com
+        {% endif %}
```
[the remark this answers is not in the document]
bencekov: done

tests/unit/test_charm.py
```diff
@@ -176,6 +197,20 @@ def test_on_pebble_ready_service_started_when_database_is_created(harness: Harne
...
+        {
+            "default_url": "http://127.0.0.1:4455",
+            "error_url": "http://127.0.0.1:4455/error",
+            "login_url": "http://127.0.0.1:4455/login",
+            "registration_url": "http://127.0.0.1:4455/registration",
+            "consent_url": "http://127.0.0.1:4455/consent",
+            "oidc_error_url": "http://127.0.0.1:4455/oidc_error",
+            "index_url": "http://127.0.0.1:4455/index",
+        },
+    )
```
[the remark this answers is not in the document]
bencekov: done according to what we discussed.

tests/unit/test_charm.py
bencekov: done

tests/unit/test_charm.py
bencekov: done

nsklikas (approved):

nsklikas (approved):

PR #52, opened 2023-04-06
Add identity schemas logic

nsklikas:
Implements the identity schema logic per the agreed kratos spec.

- Handling schemas in the shared folder is not yet implemented (we should look into that in the future).
- Having configurable jsonnet files will be implemented through the `kratos-external-idp-integrator`.


config.yaml
```diff
@@ -18,3 +18,9 @@ options:
     description: Kratos UI URL
     type: string
     default: "http://127.0.0.1:4455/"
+  identity_schemas:
+    description: A mapping of schema_id to identity schemas.
```
gruyaume, line 22: Maybe provide an example here in the description and/or in the README of how this mapping should look like.

identity_schemas/admin_v0.json
```diff
@@ -0,0 +1,42 @@
...
+              "via": "email"
+            }
+          }
+        },
+        "name": {
+          "type": "string"
+        },
+        "phone_number": {
+          "type": "string",
+          "title": "Ti mana sou"
```
gruyaume, line 36: You sure about this one?

src/charm.py
```diff
@@ -198,6 +216,68 @@ def _get_hydra_endpoint_info(self) -> Optional[str]:
...
+                return None
+
+            try:
+                schemas = json.loads(identity_schemas)
+            except json.JSONDecodeError as e:
+                logger.error(f"identity_schemas is not a valid json: {e}")
+                logger.warning("Ignoring `identity_schemas` configuration")
+                return None
+
+            # We order the dict items to make sure the dict is ordered always has the same order
```
gruyaume, line 236: Why do we care about a dict being ordered?

src/charm.py
```diff
@@ -198,6 +216,68 @@ def _get_hydra_endpoint_info(self) -> Optional[str]:
...
+            returned_schemas = {
```
gruyaume, line 237: I'd avoid the `returned` in variable names. This also applies elsewhere in this PR.

src/charm.py
```diff
@@ -198,6 +216,68 @@ def _get_hydra_endpoint_info(self) -> Optional[str]:
 
         return oauth2_provider_url
 
+    def _get_juju_config_identity_schema_config_(self) -> Optional[Tuple[str, Dict]]:
```
gruyaume, line 219: Why have those methods return tuples instead of keeping them scoped to 1 juju config?

src/charm.py
```diff
@@ -198,6 +216,68 @@ def _get_hydra_endpoint_info(self) -> Optional[str]:
...
+
+            # We order the dict items to make sure the dict is ordered always has the same order
+            returned_schemas = {
+                schema_id: f"base64://{base64.b64encode(json.dumps(schema).encode()).decode()}"
+                for schema_id, schema in sorted(schemas.items())
+            }
+            return default_schema, returned_schemas
+        return None
+
+    def _get_shared_identity_schema_config(self) -> Optional[Tuple[str, Dict]]:
```
gruyaume, line 244: Why did we half implemented this?

src/charm.py
```diff
@@ -198,6 +216,68 @@ def _get_hydra_endpoint_info(self) -> Optional[str]:
...
+            logger.info("No shared storage attached")
+            return
+
+        path = storage[0].location
+        if list(Path(path).glob("*.json")) != []:
+            raise NotImplementedError()
+        return None
+
+    def _get_default_identity_schema_config(self) -> Tuple[Optional[str], Optional[Dict]]:
+        logger.info("Pushing default identity schemas to container")
```
gruyaume, line 256: I don't see this here

src/charm.py
```diff
@@ -198,6 +216,68 @@ def _get_hydra_endpoint_info(self) -> Optional[str]:
...
+        # We order the glob to make sure the dict is ordered always has the same order
```
gruyaume, line 257: Again, why caring about dict ordering

src/charm.py
```diff
@@ -198,6 +216,68 @@ def _get_hydra_endpoint_info(self) -> Optional[str]:
...
+            self._identity_schemas_local_dir_path / DEFAULT_SCHEMA_ID_FILE_NAME
+        )
+        with open(default_schema_id_file) as f:
+            default_schema = f.read()
+        if default_schema not in returned_schemas:
+            self.unit.status = BlockedStatus(f"Default schema `{default_schema}` can't be found")
+            return None, None
+        return default_schema, returned_schemas
+
+    def _get_identity_schema_config(self) -> Optional[Tuple[str, Dict]]:
```
gruyaume, line 272: This whole section about identity schema config is non-obvious. In which cases would we use the `shared` config for example? Consider adding a docstring here to help understand the "why" behind all of it.

src/charm.py
```diff
@@ -198,6 +216,68 @@ def _get_hydra_endpoint_info(self) -> Optional[str]:
...
+            schema_file.stem: f"file://{self._identity_schemas_default_dir_path / schema_file.name}"
+            for schema_file in sorted(self._identity_schemas_local_dir_path.glob("*.json"))
+        }
+        default_schema_id_file = (
+            self._identity_schemas_local_dir_path / DEFAULT_SCHEMA_ID_FILE_NAME
+        )
+        with open(default_schema_id_file) as f:
+            default_schema = f.read()
+        if default_schema not in returned_schemas:
+            self.unit.status = BlockedStatus(f"Default schema `{default_schema}` can't be found")
```
gruyaume, line 268: Why are we changing thte juju status in this method. It seems like it should be out of its scope.

gruyaume (changes requested):

src/charm.py
gruyaume: I'd consider not implementing it at all or actually doing it but going halfway brings no value here.

nsklikas: I didn't want to implement it yet since we will need this when we have the admin UI, but I guess it does not make sense to have it as a placeholder.

I will try to provide a basic implementation and we can change it when the time comes if needed.

src/charm.py
nsklikas: Removed

src/charm.py
nsklikas: Added a small docstring explaining the logic

src/charm.py
nsklikas: I changed it throw instead

src/charm.py
```diff
@@ -239,6 +253,83 @@ def _get_hydra_endpoint_info(self) -> Optional[str]:
...
+        if default_schema not in schemas:
+            raise RuntimeError(f"Default schema `{default_schema}` can't be found")
+        return default_schema, schemas
+
+    def _get_identity_schema_config(self) -> Optional[Tuple[str, Dict]]:
+        """Get the the default schema id and the identity schemas.
+
+        The identity schemas can come from various sources. We chose them in this order:
+        1) If the user has defined some schemas in the juju config, return those
+        2) Else if there are identity schemas in the shared folder that were
```
gruyaume, line 321: Is there a reason/usecase why we want this option 2?

src/charm.py
gruyaume: Even if it wasn't, would we care? If yes maybe we should use a different data structure

src/charm.py
nsklikas: Kratos offers no API for updating the identity schemas, this is how we decided to overcome this for the time being:
- The admin login UI will place the schemas provided from the user in the shared volume
- The kratos charm will scan the shared folder and update the kratos config if any schema is found in the shared volume.

gruyaume: The shared volume is shared by the login UI and Kratos operators? If that's the case why not use relations?

nsklikas: The admin UI (which does not exist yet) and the Kratos container. We can't use relations because we want to the user to define some schema in the admin UI and then push that schema in the kratos container either in the form of a file or base64 encoded in the the kratos config.

If we were to use relations for this we would have to have a flow like the following:
- User defines schema in the admin UI and store it somewhere.
- The admin UI charm calls some API of the admin UI to fetch the schemas and updates the relation databag.
- The kratos charm listens for changes in the relation databag and updates the kratos config.

This flow seems more complicated and we have to keep 3 different storages in sync (the admin UI storage, the relation databag and the kratos storage).

Also both solutions suffer from the same problem that we can't trigger the kratos operator to update its config once a schema is altered in the admin UI. (Perhaps we could use a pebble check for this?)

We know that this approach is not optimal and problems may arise on the way but implementing the admin ui is far in the future so we can re-iterate on this when the time comes. Any suggestions are more than welcome.

gruyaume: All right, I'll resolve the comment, is this a spec somewhere?

nsklikas: There is a brief description on https://docs.google.com/document/d/13LKn-LmFom1ahUNzVFz7eU3G2aEQN4FpwuZxhB8QM8Q/edit#heading=h.q48gjgxr36ad. There is not much detail there yet.

claim_mappers/default_schema.jsonnet
```diff
@@ -21,7 +21,7 @@ local claims = {
       [if 'zoneinfo' in claims then 'zoneinfo' else null]: claims.zoneinfo,
       [if 'locale' in claims then 'locale' else null]: claims.locale,
       [if 'phone_number' in claims && claims.phone_number_verified then 'phone_number' else null]: claims.phone_number,
-      [if 'locale' in claims then 'locale' else null]: claims.locale,
+      [if 'locale' in claims then 'locale' else null]: claims.team,
```
natalian98, line 24:
```suggestion
      [if 'locale' in claims then 'locale' else null]: claims.locale,
```

src/charm.py
gruyaume: I guess I'm still unsure as to why we implemented this here if we're not sure how/when this will be set up. If we know we won't need it right now and there's no way to test it, I'd wait it out.

config.yaml
```diff
@@ -18,3 +18,40 @@ options:
...
+                  "type": "string"
+                }
+              }
+            },
+            "additionalProperties": true
+          }
+        }
+      }
+    type: string
+  default_identity_schema:
```
natalian98, line 55: It's up to you, but maybe `default_identity_schema_id` would be more accurate.

natalian98 (dismissed):
lgtm, just one more minor suggestion

src/charm.py
nsklikas: I have no problem with that, I will remove it for now.

natalian98 (approved):

gruyaume (approved):

PR #56, opened 2023-04-19
Add network policies

nsklikas:
Apply network policies to deny ingress traffic to the Kratos APIs when it isn't coming from the traefik ingress.


lib/charms/kratos/v0/kubernetes_network_policies.py
```diff
@@ -0,0 +1,160 @@
...
+    def __init__(self, charm: CharmBase) -> None:
+        self._charm = charm
+        self.client = Client(field_manager=charm.app.name, namespace=charm.model.name)
+
+    @property
+    def policy_name(self) -> str:
+        """The default policy name that will be created."""
+        return f"{self._charm.app.name}-network-policy"
+
+    def apply_ingress_policy(
```
gruyaume, line 85: We expect a list of policies so we should probably call the method `apply_ingress_policies`

lib/charms/kratos/v0/kubernetes_network_policies.py
```diff
@@ -0,0 +1,160 @@
...
+            )
+        except ApiError as e:
+            if e.status.code == 403:
+                msg = f"Kubernetes resources patch failed: `juju trust` this application. {e}"
+            else:
+                msg = f"Kubernetes resources patch failed: {e}"
+            logger.error(msg)
+            raise NetworkPoliciesHandlerError()
+
+    def delete_network_policy(self, name: Optional[str] = None) -> None:
```
gruyaume, line 147: We should standardise on naming, we use `ingress policies` and `network policies` interchangeably.

src/charm.py
```diff
@@ -522,6 +536,22 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
+            self.network_policy_handler.apply_ingress_policy(
+                [
+                    (KRATOS_PUBLIC_PORT, [self.public_ingress.relation]),
+                    (KRATOS_ADMIN_PORT, [self.admin_ingress.relation]),
+                    (38812, []),
+                    (38813, []),
+                ]
+            )
+        except NetworkPoliciesHandlerError:
+            event.defer()
```
gruyaume, line 553: Shouldn't we simply change the status to blocked (or not catch the error and let the charm go to error state)? If something is broken we may not want to be deferring forever

lib/charms/kratos/v0/kubernetes_network_policies.py
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/env python3
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Interface library for creating network policies.
```
gruyaume, line 5: Not really an "Interface" library

lib/charms/kratos/v0/kubernetes_network_policies.py
```diff
@@ -0,0 +1,160 @@
...
+```shell
+cd some-charm
+charmcraft fetch-lib charms.kratos.v0.kubernetes_network_policies
+```
+Then, to initialise the library:
+```python
+from charms.kratos.v0.kubernetes_network_policies import (
+    K8sNetworkPoliciesHandler,
+    NetworkPoliciesHandlerError,
+    PortDefinition,
```
gruyaume, line 18: This doesn't show up in the `kubernetes_network_policies` file

lib/charms/kratos/v0/kubernetes_network_policies.py
```diff
@@ -0,0 +1,160 @@
...
+
+        policy = NetworkPolicy(
+            metadata=ObjectMeta(name=name),
+            spec=NetworkPolicySpec(
+                podSelector=LabelSelector(
+                    matchLabels={
+                        "app.kubernetes.io/name": self._charm.app.name,
+                    }
+                ),
+                policyTypes=["Ingress", "Egress"],
```
gruyaume, line 129: Why have both "Ingress" and "Egress" but ask users to only provide the "ingress" policy

nsklikas: I had some troubles with dns and I was trying out things to figure it out, I should have changed the status to draft (sorry about that).

It is fixed now (I think it was some issue with my microk8s cluster that wasn't fixed by `reset`, but is fixed after `snap purge`).

gruyaume (changes requested):

src/charm.py
```diff
@@ -442,6 +449,9 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
         """Event Handler for config changed event."""
         self._handle_status_update_config(event)
 
+    def _cleanup(self, event: RemoveEvent) -> None:
+        self.network_policy_handler.delete_ingress_policies()
```
natalian98, line 453: It might be a juju bug, but when I try removing kratos it gets into terminated state with unit lost. When I use `--force --no-wait`, the charm is removed but in both cases the network policy is not deleted.

lib/charms/kratos/v0/kubernetes_network_policies.py
```diff
@@ -0,0 +1,159 @@
...
+
+    @property
+    def policy_name(self) -> str:
+        """The default policy name that will be created."""
+        return f"{self._charm.app.name}-network-policy"
+
+    def apply_ingress_policies(
+        self, policies: List[IngressPolicyDefinition], name: Optional[str] = None
+    ) -> None:
+        """Apply an ingress network policy about a related application.
```
natalian98, line 87:
```suggestion
        """Apply an ingress network policy for a related application.
```

tests/integration/conftest.py
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env python3
+# Copyright 2022 Canonical Ltd.
```
natalian98, line 2:
```suggestion
# Copyright 2023 Canonical Ltd.
```

src/charm.py
nsklikas: So I tried testing it a little more and when I test it manually, the policy is removed. I wrote an integration test to validate this and in the test the policy is not removed. This is very weird.

nsklikas: As far as I can tell the stop event is never fired, this is most likely a juju bug and not something we can control. Let's leave this as is for now since it is the best we can do and juju will (hopefully) fix it in a later version.

nsklikas:
In case applying the network policies fails we have several options:
1) Log the error and keep the normal operation, if the user wants they can deploy the charm (or re trigger the hook)
2) Go to a blocked state, this would make state management much more complicated
3) Raise an exception 
4) Log an error and defer the event until patching succeeds

Since it is not needed for the charm operation and it is not essential for security I think that (1) and (4) make most sense because they allow the charm to run despite any error that might occur and the user can make the decision to stop the charm if they deem it necessary.

If we decide to go with (1) we can also provide an action for applying the network policies to allow the user to retry applying them if something goes wrong. cc @natalian98 @gruyaume

PR #60, opened 2023-05-10
IAM-207-Kratos-COS Integration

bencekov:
Integrating Kratos charm with COS components: Prometheus and Loki

cos integrations manual testing instructions:

These instructions presuppose that you have a juju cluster with working kratos, prometheus-k8s, and loki-k8s integrated with kratos. Kratos charm uses the default relation names.

Prometheus:
curl 'prometheus_url:9090/api/v1/query?query=http_requests_total&app=kratos&code=200&endpoint=/admin/health/alive' | yq
If you get a list of metrics, then it worked.

Loki
curl loki_url:3100/loki/api/v1/label/juju_unit/values
If your unit shows up in the results, it should be good.


src/charm.py
```diff
@@ -137,6 +148,28 @@ def __init__(self, *args: Any) -> None:
...
+                        }
+                    ],
+                }
+            ],
+        )
+
+        self.loki_consumer = LogProxyConsumer(
+            self,
+            log_files=[self._log_path],
+            relation_name=self._loki_push_api_relation_name,
```
sed-i, line 169: **Nitpick:** You're using the default relation name in metadata ("logging", which is great!) so you don't _have_ to provided it here.

src/charm.py
```diff
@@ -723,6 +775,42 @@ def _on_run_migration_action(self, event: ActionEvent) -> None:
...
+        log_file_info_list = self._container.list_files(path=self._log_path)
+
+        if not log_file_info_list:
+            logger.info(
+                "Log file has not been created. Log file check delayed to next status update."
+            )
+            return
+
+        if int(log_file_info_list[0].size) > self.config["log_max_size"] * 1000000:
+            logger.info("Log file greater than allowed size. Redeploying container.")
```
sed-i, line 810: **Question:** What do you mean by "redeploying container"?

Perhaps [head-tail](https://unix.stackexchange.com/questions/367055/how-do-i-trim-bytes-from-the-beginning-and-end-of-a-file) or [RotatingFileHandler](https://stackoverflow.com/questions/9106795/python-logging-and-rotating-files) could be useful?

src/charm.py
```diff
@@ -197,7 +247,9 @@ def _pebble_layer(self) -> Layer:
                     "override": "replace",
                     "summary": "Kratos Operator layer",
                     "startup": "disabled",
-                    "command": "kratos serve all " + self._kratos_service_params,
+                    "command": '/bin/sh -c "{} {} 2>&1 | tee -a {}"'.format(
```
gruyaume, line 250: Why add `/bin/sh -c`?

src/charm.py
```diff
@@ -723,6 +775,42 @@ def _on_run_migration_action(self, event: ActionEvent) -> None:
             return
         event.log("Successfully migrated the database.")
 
+    def _on_loki_consumer_endpoint_joined(self, event: LogProxyEndpointJoined) -> None:
+        logger.info("Loki Push API endpoint joined")
+
+    def _on_loki_consumer_endpoint_departed(self, event: LogProxyEndpointDeparted) -> None:
```
gruyaume, line 781: Similar to @sed-i 's comment above, do we need to handle this event?

bencekov: Not really, I left it in for logging purpuses. It was useful while debugging. I'll take it out then.

src/charm.py
```diff
@@ -723,6 +775,42 @@ def _on_run_migration_action(self, event: ActionEvent) -> None:
             return
         event.log("Successfully migrated the database.")
 
+    def _on_loki_consumer_endpoint_joined(self, event: LogProxyEndpointJoined) -> None:
+        logger.info("Loki Push API endpoint joined")
```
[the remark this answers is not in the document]
bencekov: I put it in for logging, and debugging purposes. It seems there's a consensus that it should be removed.

src/charm.py
```diff
@@ -98,6 +105,10 @@ def __init__(self, *args: Any) -> None:
         self._db_relation_name = "pg-database"
         self._hydra_relation_name = "endpoint-info"
         self._login_ui_relation_name = "ui-endpoint-info"
+        self._prometheus_scrape_relation_name = "metrics-endpoint"
+        self._loki_push_api_relation_name = "logging"
```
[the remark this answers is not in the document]
bencekov: I see your point, but for now I rather stay consistent with the rest of the charm, and other charms in our bundle.

src/charm.py
bencekov: It's consistent with the rest of the charm, and I prefer it because the docs in loki were a bit misleading/confusing: I specifically mean the loki_push_api lib line 46.

src/charm.py
```diff
@@ -723,6 +775,42 @@ def _on_run_migration_action(self, event: ActionEvent) -> None:
...
+        self.unit.status = BlockedStatus(event.message)
+
+    def _on_update_status_handle_logging(self, event: HookEvent) -> None:
+        if not self._container.can_connect():
+            logger.info(
+                "Cannot connect to Kratos container. Log file check delayed to next status update."
+            )
+            return
+
+        if not self.unit.status == ActiveStatus():
```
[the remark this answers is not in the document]
bencekov: Cool idea, but this is a refactor for the whole charm. (Also our other charms in the bundle.) I try to keep my PRs to focus on specific issues.

src/charm.py
```diff
@@ -723,6 +760,35 @@ def _on_run_migration_action(self, event: ActionEvent) -> None:
...
+
+        if not log_file_info_list:
+            logger.info(
+                "Log file has not been created. Log file check delayed to next status update."
+            )
+            return
+
+        if int(log_file_info_list[0].size) > int(self.config["log_max_size"]) * 1000000:
+            logger.info("Log file greater than allowed size. Restarting container.")
+            self._handle_status_update_config(event)
```
nsklikas, line 790: This function does not restart the container. It can be confusing because the same variable is used for the container and for the service, but this function will only restart the service. I don't think you can restart the container through pebble. 

It is not easy to handle file rotation (you would have to move the file -> restart kratos -> delete the old file after a while?, in this case we would have to provide the whole folder to loki?). Perhaps it is best to let the container die when the disk limit is reached, unless we can provide a solution that we know works well and in all cases.

nsklikas:
I tried running this and I have a couple of issues/questions:
- The log rotation does not work
- To access the logs I used the grafana dashboard. I tried to get all the logs and it seems that the labels are a little bit malformed, is this some configuration that we need to change or am I missing something?
image
- From what I understand promtail is used to push the logs to loki. How does that work, does the promtail binary have to be installed in the container? I don't understand where that happens in you changes.

src/charm.py
bencekov: I realize since that this function only restarts the service, not the container. I'm not sure how I can use either solution without introducing some problems with filehandlers or a race condition for the log file, therefore for now I'll remove the log rotation functionality all together.

src/charm.py
bencekov: so for now I'll remove the logrotation, and raise it as an issue.

src/charm.py
sed-i: Haven't checked, but how about
1. stop service
2. trim log
3. start service

BTW what does kratos do with logs? Just let them grow?

nsklikas: Kratos just writes to stdout, I suppose in that case log rotation is managed by kubelet. We discussed with @bencekov a similar approach to the one you described (not trimming the logs, but rotating the log file). 

Tbh I am not familiar with promtail and I don't know how it would behave if we trimmed logs from the file, but it sounds error prone. Do you think this is the best way forward or are there any other projects that have implemented this approach?

sed-i: Ah right.
I think promtail is able to pick up a head-trim, but @rbarry82 can probably advise more accurately.
Grafana labs [recommend](https://grafana.com/docs/loki/latest/clients/promtail/logrotation/#rename-and-create) using "rename and create".

config.yaml
```diff
@@ -51,3 +51,7 @@ options:
   default_identity_schema_id:
     description: The default identity schema id, this option only works when `identity_schemas` is defined
     type: string
+  log_max_size:
```
nsklikas, line 54: This is no longer needed

tests/unit/test_charm.py
```diff
@@ -1346,3 +1380,31 @@ def test_timeout_on_run_migration(
     harness.charm._on_run_migration_action(event)
 
     event.fail.assert_called()
+
+
+def test_on_pebble_ready_with_loki_(harness: Harness) -> None:
```
nsklikas, line 1385:
```suggestion
def test_on_pebble_ready_with_loki(harness: Harness) -> None:
```

src/charm.py
```diff
@@ -137,6 +143,28 @@ def __init__(self, *args: Any) -> None:
...
 
+        self.metrics_endpoint = MetricsEndpointProvider(
+            self,
+            relation_name=self._prometheus_scrape_relation_name,
+            jobs=[
+                {
+                    "metrics_path": "/metrics/prometheus",
+                    "static_configs": [
+                        {
+                            "targets": ["*:4434"],
```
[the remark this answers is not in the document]
bencekov: I verified this manually

src/charm.py
```diff
@@ -98,6 +102,10 @@ def __init__(self, *args: Any) -> None:
         self._db_relation_name = "pg-database"
         self._hydra_relation_name = "endpoint-info"
         self._login_ui_relation_name = "ui-endpoint-info"
+        self._prometheus_scrape_relation_name = "metrics-endpoint"
+        self._loki_push_api_relation_name = "logging"
+        self._kratos_service_command = "kratos serve all"
```
nsklikas, line 107: Why move this to a variable?

src/charm.py
```diff
@@ -217,6 +254,13 @@ def _pebble_layer(self) -> Layer:
     def _domain_url(self) -> Optional[str]:
         return self.config["external_url"] or self.public_ingress.url
 
+    @property
+    def _log_level(self) -> str:
+        level = self.config["log_level"]
+        if level not in LOG_LEVELS:
+            return DEFAULT_LOG_LEVEL
```
nsklikas, line 261: Not sure if we want this. At the very least if the user has provided an incorrect log level we need to warn them about it.

nsklikas (commented):
LGTM, left 2 very minor nitpicks.

src/charm.py
bencekov: relation names are variables, also with loki there's a confusion about what the default relation name is. Also I assumed it's better to use a variable string then to hardcode the command into the pebble layer.

bencekov: also this way the charm code is more uniform across the bundle

config.yaml
```diff
@@ -51,3 +51,9 @@ options:
   default_identity_schema_id:
     description: The default identity schema id, this option only works when `identity_schemas` is defined
     type: string
+  log_level:
+    description: |
+      The value controlling the verbosity of logs produced by Kratos.
```
gruyaume, line 56:
```suggestion
      The verbosity of logs produced by Kratos.
```

src/charm.py
```diff
@@ -137,6 +145,28 @@ def __init__(self, *args: Any) -> None:
...
 
+        self.metrics_endpoint = MetricsEndpointProvider(
+            self,
+            relation_name=self._prometheus_scrape_relation_name,
+            jobs=[
+                {
+                    "metrics_path": "/metrics/prometheus",
+                    "static_configs": [
+                        {
+                            "targets": ["*:4434"],
```
gruyaume, line 156: There's already a variable for `4434` right, we probably want to leverage it. The same applies at a couple of places in this file.

src/charm.py
gruyaume: Agreed, it's fine to go in blocked state if the user config is not correct. We don't want to encourage/tolerate writing bad configurations.

gruyaume (changes requested):

src/charm.py
```diff
@@ -80,6 +83,19 @@ def dict_to_action_output(d: Dict) -> Dict:
...
+    """This decorator puts the charm in Blocked State in log_level is misconfigured."""
+
+    @wraps(method)
+    def config_check(self, *args, **kwargs):
+        if not self._log_level_valid:
+            self.unit.status = BlockedStatus("Bad configuration value for log_level")
+            return
+        return method(self, *args, **kwargs)
+
+    return config_check
```
nsklikas, line 96: I don't like this approach. It's counter-intuitive to use a decorator to validate a class property. Correct me if I'm wrong but there is only one place where we need to validate the config and that's the `on.config_changed` hook. I suppose that your problem is that other hooks may overwrite the unit state. This is an indication that we handle status setting badly and that we need to refactor it. Having only one function that sets the unit's state is probably a step in the right direction.

gruyaume (approved):

natalian98:
Could you update the CI workflow file with the new juju command to get kratos unit logs, so that we can see in details what fails in the integration tests? Like [here](https://github.com/canonical/hydra-operator/blob/main/.github/workflows/tests.yaml#L50)

src/charm.py
natalian98: What's the confusion about the default loki relation name about? Isn't it just [logging](https://github.com/canonical/kratos-operator/blob/9835be701fccf3ef8454f993b89f7729da09eac9/lib/charms/loki_k8s/v0/loki_push_api.py#L491)?

src/charm.py
```diff
@@ -421,6 +471,11 @@ def _on_pebble_ready(self, event: PebbleReadyEvent) -> None:
         logger.info("Pebble plan updated with new configuration, replanning")
         self._container.replan()
 
+        if not self._log_level_valid:
+            logger.info(f"Invalid configuration value for log_level: {self._log_level}")
+            self.unit.status = BlockedStatus("Bad configuration value for log_level")
+            return
```
nsklikas, line 477: This could be a function to avoid code duplication (`_validate_config`?). Shouldn't this event (and some of the other events) be deferred if the configuration is bad?

src/charm.py
```diff
@@ -473,6 +528,11 @@ def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
             logger.info("Kratos service is absent. Deferring database created event.")
             return
 
+        if not self._log_level_valid:
+            logger.info(f"Invalid configuration value for log_level: {self._log_level}")
+            self.unit.status = BlockedStatus("Bad configuration value for log_level")
+            return
```
nsklikas, line 534: Shouldn't this event be deferred if the configuration is bad?

nsklikas (changes requested):

src/charm.py
```diff
@@ -536,6 +594,10 @@ def _on_client_config_changed(self, event: ClientConfigChangedEvent) -> None:
             event.defer()
             return
 
+        if not self._validate_config_log_level():
+            event.defer()
+            return
```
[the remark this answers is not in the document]
bencekov: I realised that this is probably not a good approach. The only relevant handlers for checking the log_level are at the start, and when there's a config update.

src/charm.py
```diff
@@ -461,6 +515,10 @@ def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
             self.unit.status = WaitingStatus("Waiting to connect to Kratos container")
             return
 
+        if not self._validate_config_log_level():
+            event.defer()
+            return
```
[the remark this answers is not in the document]
bencekov: I realised that this is probably not a good approach. The only relevant handlers for checking the log_level are at the start, and when there's a config update.

tests/unit/test_charm.py
```diff
@@ -1328,3 +1389,41 @@ def test_timeout_on_run_migration(
...
+
+
+def test_on_pebble_ready_with_loki(harness: Harness) -> None:
+    setup_postgres_relation(harness)
+    setup_peer_relation(harness)
+    container = harness.model.unit.get_container(CONTAINER_NAME)
+    harness.charm.on.kratos_pebble_ready.emit(container)
+    setup_loki_relation(harness)
+
+    assert harness.model.unit.status == ActiveStatus()
```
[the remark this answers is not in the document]
bencekov: hey I mocked out the lib call. How do I find out how long the unit test takes to complete?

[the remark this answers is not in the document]
nsklikas: You run it locally, not sure if you can do it from the github API

nsklikas (dismissed):

nsklikas (approved):
LGTM, one very minor comment and I think we can merge it

tests/unit/conftest.py
```diff
@@ -202,3 +203,11 @@ def mocked_create_identity(mocker: MockerFixture, kratos_identity_json: Dict) ->
 def mocked_run_migration(mocker: MockerFixture) -> MagicMock:
     mock = mocker.patch("charm.KratosAPI.run_migration", return_value=(None, None))
     return mock
+
+
+@pytest.fixture()
```
nsklikas, line 208: Perhaps it would make sense to set the `autouse=True` flag, so that this is mocked in all tests to avoid fetching the promtail binary by accident

nsklikas (dismissed):
LGTM, one very minor comment and I think we can merge it

src/charm.py
```diff
@@ -377,6 +425,9 @@ def _handle_status_update_config(self, event: HookEvent) -> None:
 
         self.unit.status = MaintenanceStatus("Configuring/deploying resources")
 
+        if not self._validate_config_log_level():
```
natalian98, line 428: Let's also move this check before calling `template.render`

gruyaume (approved):

natalian98 (approved):

PR #67, opened 2023-05-17
Use ROCK image

nsklikas:
Switch metadata image to canonical ROCK

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -53,5 +53,12 @@ jobs:
       - name: Get kratos container logs
         run: kubectl logs kratos-0 -c kratos -n testing
 
+      # Hack to overcome lack of tools (cat, tar) in the workload container
       - name: Get kratos config file
-        run: kubectl exec -it kratos-0 -c kratos -n testing -- cat /etc/config/kratos.yaml
+        run: |
+          juju ssh kratos/0 "PYTHONPATH=agents/unit-kratos-0/charm/venv/ python3 -c '
+          from ops import pebble
+          p = pebble.Client(\"/charm/containers/kratos/pebble.socket\")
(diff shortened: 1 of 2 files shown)


metadata.yaml
gruyaume, line 1: Integration tests don't pass because we're using `cat` there.

gruyaume (changes requested):

.github/workflows/tests.yaml
```diff
@@ -53,5 +53,12 @@ jobs:
...
+      # Hack to overcome lack of tools (cat, tar) in the workload container
       - name: Get kratos config file
-        run: kubectl exec -it kratos-0 -c kratos -n testing -- cat /etc/config/kratos.yaml
+        run: |
+          juju ssh kratos/0 "PYTHONPATH=agents/unit-kratos-0/charm/venv/ python3 -c '
+          from ops import pebble
+          p = pebble.Client(\"/charm/containers/kratos/pebble.socket\")
+          f = p.pull(\"/etc/config/kratos.yaml\")
+          print(f.read())
+          '"
```
gruyaume, line 64: Is looking for the config file an appropriate test for the rock? I'd do something that validates that the binary works. Anyway this is not a related to this PR since the integration tests were already doing this.

gruyaume (approved):

.github/workflows/tests.yaml
nsklikas: I suppose it is there for debugging purposes, if something goes wrong then this would help debug it. Having said that, you should be able to get the same config file if you run the integration tests locally, so I am not opposed to removing this step.

nsklikas: I suppose this part is missing:
```
        if: failure()
```

PR #70, opened 2023-05-24
 Convert ingress provided URLs to https

nsklikas:
- Convert ingress provided URLs to https and remove the port.
- Remove `external_url` config.

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -44,6 +44,7 @@ def normalise_url(url: str) -> str:
     1) The ingress will serve https under the 443 port, the user-agent will
        implicitly make the request on that port
     2) The provided URL is not a relative path
+    3) No user/password is provided in the netloc
 
     This is a hack and should be removed once traefik provides a way for us to
     request the https URL.
--- a/src/utils.py
+++ b/src/utils.py
@@ -9,7 +9,16 @@ from urllib.parse import urlparse, urlunparse
 
 
 def dict_to_action_output(d: Dict) -> Dict:
-    """Convert all keys in a dict to the format of a juju action output."""
+    """Convert all keys in a dict to the format of a juju action output.
+
+    All `_` in the keys are replaced with `-`. This is applied recursively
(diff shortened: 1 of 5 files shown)


src/utils.py
```diff
@@ -0,0 +1,42 @@
...
+# See LICENSE file for licensing details.
+
+"""Utility functions for the Kratos charm."""
+
+from typing import Dict
+from urllib.parse import urlparse, urlunparse
+
+
+def dict_to_action_output(d: Dict) -> Dict:
+    """Convert all keys in a dict to the format of a juju action output."""
```
gruyaume, line 12: Can you please add an example in the docstring to make it clearer what this method does?

src/utils.py
```diff
@@ -0,0 +1,42 @@
...
+    for k, v in d.items():
+        k = k.replace("_", "-")
+        if isinstance(v, dict):
+            v = dict_to_action_output(v)
+        ret[k] = v
+    return ret
+
+
+def normalise_url(url: str) -> str:
+    """Convert a URL to a more userfriendly format.
```
gruyaume, line 23: Can you please add an example in the docstring to make it clearer what this method does?

src/utils.py
```diff
@@ -0,0 +1,42 @@
...
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Utility functions for the Kratos charm."""
+
+from typing import Dict
+from urllib.parse import urlparse, urlunparse
+
+
+def dict_to_action_output(d: Dict) -> Dict:
```
gruyaume, line 11: You should have unit tests specifically for those new methods.

nsklikas: Done

src/utils.py
nsklikas: Done

src/utils.py
nsklikas: Done

PR #80, opened 2023-06-06
fix: fixed issue around the log file directory

bencekov:
Fixed the issue of directory in log file path not being present in rock. Manually verified to work.

To test:
- deploy kratos
- deploy loki
- integrate kratos with loki
- after entering ```curl loki_url:3100/loki/api/v1/label/juju_unit/values``` command you should see the kratos unit you deployed

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -515,11 +515,11 @@ class KratosCharm(CharmBase):
         # Necessary directory for log forwarding
         if not self._container.can_connect():
             event.defer()
-            logger.info("Cannot connect to Hydra container. Deferring event.")
             self.unit.status = WaitingStatus("Waiting to connect to Kratos container")
             return
         if not self._container.isdir(str(self._log_dir)):
-            self._container.make_dir(path=str(self._log_dir), make_parents=True, permissions=0o755)
+            self._container.make_dir(path=str(self._log_dir), make_parents=True)
+            logger.info(f"Created directory {self._log_dir}")
 
         self._handle_status_update_config(event)
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -107,8 +107,8 @@ class KratosCharm(CharmBase):
         self._prometheus_scrape_relation_name = "metrics-endpoint"
         self._loki_push_api_relation_name = "logging"
         self._kratos_service_command = "kratos serve all"
-        self._log_path = "/var/log/kratos.log"
         self._log_dir = "/var/log"
+        self._log_path = f"{self._log_dir}/kratos.log"
 
         self.kratos = KratosAPI(
             f"http://127.0.0.1:{KRATOS_ADMIN_PORT}", self._container, str(self._config_file_path)
--- a/src/charm.py
+++ b/src/charm.py
@@ -500,8 +500,6 @@ class KratosCharm(CharmBase):
             return
 
         self._push_default_files()
-        if not self._container.isdir(self._log_dir):
-            self._container.make_dir(path=self._log_dir, make_parents=True, permissions=0o777)
 
     def _on_leader_elected(self, event: LeaderElectedEvent) -> None:
         if not self.unit.is_leader():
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -463,6 +464,8 @@ def _on_install(self, event: InstallEvent) -> None:
             return
 
         self._push_default_files()
+        if not self._container.isdir(self._log_dir):
```
nsklikas, line 467: Do we need to check if the dir exists? Can't we just call the `make_dir` function?

nsklikas (commented):
Can you add a couple of tests for this change?

src/charm.py
bencekov: I haven't verified that if an image has this dir already, the function doesn't error out. So I checked for it to not be there.

gruyaume (approved):

src/charm.py
```diff
@@ -476,7 +476,7 @@ def _on_pebble_ready(self, event: PebbleReadyEvent) -> None:
             self.unit.status = WaitingStatus("Waiting to connect to Kratos container")
             return
         if not self._container.isdir(str(self._log_dir)):
-            self._container.make_dir(path=str(self._log_dir), make_parents=True, permissions=0o777)
+            self._container.make_dir(path=str(self._log_dir), make_parents=True, permissions=0o755)
```
natalian98, line 479: It's the [default](https://ops.readthedocs.io/en/latest/#ops.Container.make_dir) pebble permission, so you don't need to specify it.

src/charm.py
```diff
@@ -95,7 +95,8 @@ def __init__(self, *args: Any) -> None:
         self._prometheus_scrape_relation_name = "metrics-endpoint"
         self._loki_push_api_relation_name = "logging"
         self._kratos_service_command = "kratos serve all"
-        self._log_path = "/var/log/kratos.log"
+        self._log_dir = Path("/var/log")
```
natalian98, line 98: Do we need to make it a `Path` object if it needs to be passed as a string afterwards?

src/charm.py
bencekov: taken out

tests/unit/test_charm.py
```diff
@@ -1284,3 +1284,25 @@ def test_on_config_changed_with_invalid_log_level(harness: Harness) -> None:
...
+
+
+def test_on_pebble_ready_make_dir_called(
+    harness: Harness, mocked_isdir: MagicMock, mocked_make_dir: MagicMock
+) -> None:
+    container = harness.model.unit.get_container(CONTAINER_NAME)
+    harness.charm.on.kratos_pebble_ready.emit(container)
+
+    mocked_isdir.assert_called_once_with("/var/log")
+    mocked_make_dir.assert_called_once_with(path="/var/log", make_parents=True)
```
nsklikas, line 1296: I think this is cleaner:
```suggestion
def test_on_pebble_ready_make_dir_called(harness: Harness) -> None:
    container = harness.model.unit.get_container(CONTAINER_NAME)
    harness.charm.on.kratos_pebble_ready.emit(container)

    assert container.isdir("/var/log")
```

tests/unit/conftest.py
```diff
@@ -211,3 +211,15 @@ def mocked_log_proxy_consumer_setup_promtail(mocker: MockerFixture) -> MagicMock
...
+@pytest.fixture()
+def mocked_isdir(mocker: MockerFixture) -> MagicMock:
+    mocked_isdir = mocker.patch("ops.Container.isdir", return_value=False)
+    return mocked_isdir
+
+
+@pytest.fixture()
+def mocked_make_dir(mocker: MockerFixture) -> MagicMock:
+    mocked_make_dir = mocker.patch("ops.Container.make_dir", return_value=None)
+    return mocked_make_dir
```
nsklikas, line 225: Harness takes care of this, no need to mock it.

nsklikas (changes requested):
Let's clean up the tests and merge this.

tests/unit/conftest.py
bencekov: done

tests/unit/test_charm.py
```diff
@@ -1284,3 +1284,25 @@ def test_on_config_changed_with_invalid_log_level(harness: Harness) -> None:
...
+def test_on_pebble_ready_cannot_connect_container_make_dir_not_called(
+    harness: Harness, mocked_isdir: MagicMock, mocked_make_dir: MagicMock
+) -> None:
+    harness.set_can_connect(CONTAINER_NAME, False)
+    container = harness.model.unit.get_container(CONTAINER_NAME)
+    harness.charm.on.kratos_pebble_ready.emit(container)
+
+    assert harness.model.unit.status == WaitingStatus("Waiting to connect to Kratos container")
+    mocked_isdir.assert_not_called
+    mocked_make_dir.assert_not_called
```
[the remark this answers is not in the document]
bencekov: done

tests/unit/test_charm.py
bencekov: done

nsklikas (approved):

PR #83, opened 2023-06-12
Fix cookie secret sharing

### LINKED ISSUE #82: Cookie secret is not shared across units
### Bug Description

The cookies secret is generated every time kratos starts, this means that each unit has a different secret and that cookies generated by one unit cannot be read by the other units.

nsklikas:
Generate cookie secret and store it in juju secrets.

Closes #82

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -60,7 +60,6 @@ from ops.model import (
     BlockedStatus,
     MaintenanceStatus,
     ModelError,
-    Relation,
     Secret,
     SecretNotFoundError,
     WaitingStatus,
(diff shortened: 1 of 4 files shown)


src/charm.py
```diff
@@ -257,6 +269,11 @@ def _get_available_mappers(self) -> List[str]:
             for schema_file in self._mappers_local_dir_path.iterdir()
         ]
 
+    @property
+    def _peers(self) -> Optional[Relation]:
```
gruyaume, line 273: The PR description claims that the secret is stored in the peer relation databag but I do not see were this is done. All I see is that the secret ends up being created.

nsklikas: You are correct. I thought that after all juju secrets are a better place to store this, but I forgot to change the description.

Still there is a weird bug when I run this code in the github runners, so I should place it to draft for now.

gruyaume (approved):

src/charm.py
```diff
@@ -444,6 +471,13 @@ def _handle_status_update_config(self, event: HookEvent) -> None:
             event.defer()
             return
 
+        if not self._get_secret():
+            self.unit.status = WaitingStatus("Waiting for secret creation")
+            if not self.unit.is_leader():
+                event.defer()
```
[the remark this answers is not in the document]
nsklikas: Makes sense, I moved it there

gruyaume (approved):

natalian98 (approved):

PR #88, opened 2023-07-03
Add login and sessions endpoints to kratos lib

natalian98:
- adds login browser and sessions endpoints to the `kratos_endpoints` library
- uses ingress (if relation provided) instead of service to communicate with other components, since oathkeeper and kratos can be in different models.

### DIFF
--- a/lib/charms/kratos/v0/kratos_endpoints.py
+++ b/lib/charms/kratos/v0/kratos_endpoints.py
@@ -42,7 +42,6 @@ from typing import Dict, Optional
 
 from ops.charm import CharmBase, RelationCreatedEvent
 from ops.framework import EventBase, EventSource, Object, ObjectEvents
-from ops.model import Application
 
 # The unique Charmhub library identifier, never change it
 LIBID = "5868b36df1c04c90b33f5e5557327162"
@@ -94,13 +93,14 @@ class KratosEndpointsProvider(Object):
             return
 
         relations = self.model.relations[self._relation_name]
+        endpoints_databag = {
+            "admin_endpoint": admin_endpoint,
+            "public_endpoint": public_endpoint,
+            "login_browser_endpoint": f"{public_endpoint}/self-service/login/browser",
+            "sessions_endpoint": f"{public_endpoint}/sessions/whoami",
+        }
         for relation in relations:
-            relation.data[self._charm.app].update(
-                {
-                    "admin_endpoint": admin_endpoint,
-                    "public_endpoint": public_endpoint,
-                }
-            )
+            relation.data[self._charm.app].update(endpoints_databag)
 
 
 class KratosEndpointsRelationError(Exception):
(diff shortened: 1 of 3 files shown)


lib/charms/kratos/v0/kratos_endpoints.py
```diff
@@ -94,13 +93,14 @@ def send_endpoint_relation_data(self, admin_endpoint: str, public_endpoint: str)
             return
 
         relations = self.model.relations[self._relation_name]
+        endpoints_databag = {
+            "admin_endpoint": admin_endpoint,
+            "public_endpoint": public_endpoint,
+            "login_browser_endpoint": f"{public_endpoint}/self-service/login/browser",
+            "sessions_endpoint": f"{public_endpoint}/sessions/whoami",
```
nsklikas, line 100: We need to be consistent, we either provide the base URL  and it is the other party's responsibility to construct the full path (current implementation) or we provide all the different endpoints (your change). I am not against this approach, but we need to follow only one.

src/charm.py
```diff
@@ -531,12 +531,16 @@ def _update_kratos_endpoints_relation_data(self, event: RelationEvent) -> None:
...
+            else f"http://{self.app.name}.{self.model.name}.svc.cluster.local:{KRATOS_ADMIN_PORT}",
         )
         public_endpoint = (
-            f"http://{self.app.name}.{self.model.name}.svc.cluster.local:{KRATOS_PUBLIC_PORT}"
+            self.public_ingress.url
+            if self.public_ingress.is_ready()
+            else f"http://{self.app.name}.{self.model.name}.svc.cluster.local:{KRATOS_PUBLIC_PORT}",
         )
-        self.endpoints_provider.send_endpoint_relation_data(admin_endpoint, public_endpoint)
+        self.endpoints_provider.send_endpoint_relation_data(admin_endpoint[0], public_endpoint[0])
```
nsklikas, line 543: Why do we need to return `[0]`?

nsklikas (changes requested):
AFAIK as long as the 2 apps are in the same cluster, you don't need to use an ingress (k8s dns names work cross namespace). You would (most of the time) need to use the ingress if the 2 apps are in different controllers. To add to that if it the other application cannot resolve the service dns name, it does not make sense to default back to the k8s service dns.

Having said that, I don't think we should change all of our routing to cater this use case. Does juju/ops provide a mechanism we can use to distinguish between normal relations. cross-model relations and cross-controller relations? How do other charms handle this?

lib/charms/kratos/v0/kratos_endpoints.py
natalian98: I think we follow the latter (see [login ui lib](https://github.com/canonical/identity-platform-login-ui-operator/blob/main/lib/charms/identity_platform_login_ui_operator/v0/login_ui_endpoints.py#L111)). I don't think the other party should construct the path, it's easier to manage in the charm that provides the endpoint (in case the api ever changes).

src/charm.py
natalian98: It's because this returns a tuple:
```python3
public_endpoint = (
	self.public_ingress.url
	if self.public_ingress.is_ready()
	else f"http://{self.app.name}.{self.model.name}.svc.cluster.local:{KRATOS_PUBLIC_PORT}",
    )
```
I can cast it to string though, so that we don't have to return `[0]`.

nsklikas: ohhh, you need to remove the trailing comma

nsklikas (approved):

Issue #84, opened 2023-06-19
Kratos fails to restart due to a missing schema file


natalian98:
### Bug Description

After kratos is restarted (for example, if the juju controller was shut down), it fails to start the unit due to missing admin schema which should be pushed to `/etc/config/schemas/default/admin_v0.json` before the container is started. The unit gets blocked with a message "Failed to restart, please consult the logs".

### To Reproduce

1. `juju deploy iam --channel edge --trust`
2. Trigger juju controller restart (for example, by restarting an ec2 instance)

### Environment

ec2 t2.medium instance
juju 3.1.2
microk8s 1.26.5
kratos rev 329

### Relevant log output

```shell
$ juju debug-log --include unit-kratos-0 --replay
WARNING "unit-kratos-0" was not recognised as a valid application name. Only applications produce logs for CAAS models application
unit-kratos-0: 13:30:37 INFO juju.cmd running containerAgent [3.1.2 7009f54bec5f51139d42d28038031fa9ec620139 gc go1.20.3]
unit-kratos-0: 13:30:37 INFO juju.cmd.containeragent.unit start "unit"
unit-kratos-0: 13:30:37 INFO juju.worker.upgradesteps upgrade steps for 3.1.2 have already been run.
unit-kratos-0: 13:30:37 INFO juju.worker.probehttpserver starting http server on 127.0.0.1:65301
unit-kratos-0: 13:30:37 INFO juju.api cannot resolve "controller-service.controller-my-controller.svc.cluster.local": lookup controller-service.controller-my-controller.svc.cluster.local: operation was canceled
unit-kratos-0: 13:30:37 INFO juju.api connection established to "wss://10.152.183.53:17070/model/e41eed8e-aa10-412a-8222-1f02c07e6d96/api"
unit-kratos-0: 13:30:37 INFO juju.worker.apicaller [e41eed] "unit-kratos-0" successfully connected to "10.152.183.53:17070"
unit-kratos-0: 13:30:37 INFO juju.api cannot resolve "controller-service.controller-my-controller.svc.cluster.local": lookup controller-service.controller-my-controller.svc.cluster.local: operation was canceled
unit-kratos-0: 13:30:37 INFO juju.api connection established to "wss://10.152.183.53:17070/model/e41eed8e-aa10-412a-8222-1f02c07e6d96/api"
[...]

PR #89, opened 2023-07-04
fix: Kratos fails to restart when cluster reboots

wood-push-melon:
### Context

This pull request tries to resolve the issue https://github.com/canonical/kratos-operator/issues/84.

### Solution

We push the default schema files mappers in the callback function for `pebble_ready` event to cover the cluster restart scenario.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -148,8 +148,6 @@ class KratosCharm(CharmBase):
 
         self.endpoints_provider = KratosEndpointsProvider(self)
 
-        self.framework.observe(self.on.install, self._on_install)
-
         self.metrics_endpoint = MetricsEndpointProvider(
             self,
             relation_name=self._prometheus_scrape_relation_name,
@@ -172,6 +170,7 @@ class KratosCharm(CharmBase):
             container_name=self._container_name,
         )
 
+        self.framework.observe(self.on.install, self._on_install)
         self.framework.observe(self.on.kratos_pebble_ready, self._on_pebble_ready)
         self.framework.observe(self.on.leader_elected, self._on_leader_elected)
         self.framework.observe(self.on.config_changed, self._on_config_changed)
@@ -497,7 +496,7 @@ class KratosCharm(CharmBase):
     def _on_install(self, event: InstallEvent) -> None:
         if not self._container.can_connect():
             event.defer()
-            logger.info("Cannot connect to Kratos container. Deferring event.")
+            logger.info("Cannot connect to Kratos container. Deferring install event.")
             self.unit.status = WaitingStatus("Waiting to connect to Kratos container")
             return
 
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -521,6 +520,7 @@ def _on_pebble_ready(self, event: PebbleReadyEvent) -> None:
             self._container.make_dir(path=str(self._log_dir), make_parents=True)
             logger.info(f"Created directory {self._log_dir}")
 
+        self._push_default_files()
```
wood-push-melon, line 523: I did not add any checks to check if the default schemas and mappers already exist or need to be updated (these files are rarely changed) in the container's file system. I guess it would be OK to simply push them in the `pebble_ready` event. Let me know if it is necessary to add the check.

tests/integration/test_charm.py
```diff
@@ -67,14 +67,16 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
...
 
-    async with ops_test.fast_forward():
-        await ops_test.model.wait_for_idle(
-            apps=[APP_NAME, POSTGRES],
-            status="active",
-            raise_on_blocked=True,
-            timeout=1000,
-        )
-        assert ops_test.model.applications[APP_NAME].units[0].workload_status == "active"
+    await ops_test.model.wait_for_idle(
```
wood-push-melon, line 70: I guess `fast_forward` does not help here, so I removed it.

nsklikas (approved):
LGTM overall. I have just one comment about the changes to the test, but if I am missing something go ahead and merge this.

About you comments:
>Do you think it is better we organize the [constants](https://github.com/canonical/kratos-operator/pull/89/files#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cL94-L111)
This is something we need to do for sure. Let's create an issue for this.

> Are you aware of any approaches to improve the charm building speed
We can check if there is a charm already packed and use that, else we can build the charm like we do now. This would improve local runs. When running on the github runner some teams are using github caches, but I don't know how much of a speed up this would give us. I have to say though that we need to be careful because using a cache MAY cause unpredictable behavior if is not properly cleaned up.

> Is there any restriction for the Python version to be used to develop and run the charm
Charms are packed and uploaded to charmhub. So a charm will always run with the same version. I _think_ that the python version is the system python of the base image. I don't think we need to worry about that because we run our (unit) tests on the same ubuntu version.

src/charm.py
nsklikas: As long as it does not raise an error if the files already exist, I am fine with this.

tests/integration/test_charm.py
```diff
@@ -67,14 +67,16 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
...
+        apps=[APP_NAME, POSTGRES],
+        status="active",
+        raise_on_blocked=True,
+        timeout=1000,
+    )
+
+    assert all(
+        unit.workload_status == "active" for unit in ops_test.model.applications[APP_NAME].units
+    )
```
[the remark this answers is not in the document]
wood-push-melon: Yep, let me remove this check.

I looked through the `wait_for_idle` implementation, I guess it would be better to specify the `wait_for_exact_units` argument in the `wait_for_idle` for this scale-up test case.

nsklikas (approved):

PR #98, opened 2023-07-24
IAM-209-Add alert rules to Kratos Operator

bencekov:
Added alert rules to Kratos. Alert rules are verifiable on the grafana html user interface.

# TESTING

To verify Kratos alerts, you will need the cos-lite bundle:
```
juju deploy cos-lite --trust
```

Pack and deploy Kratos (in operator root directory), and its backend:

```
juju deploy postgresql-k8s --trust
charmcraft pack
juju deploy ./kratos*.charm  --resource oci-image=$(yq eval '.resources.oci-image.upstream-source' metadata.yaml) --trust
juju integrate kratos postgresql-k8s
```

# Integrations:

You need to integrate kratos with prometheus, loki:

```
juju integrate kratos prometheus
juju integrate kratos loki
```

Access Grafana's UI with your browser. The url is grafana-ip:3000. Username is admin, you can get the password with `juju run grafana/0 get-admin-password`

On the Grafana interface you should be able to find options bar on the left side of the screen. Go to alerts. There you will see the available alerts from prometheus and loki. You can open them up with the arrow on the left to see more details, or to open a graph mapping the result of the alert query over time.

### DIFF
--- a/src/loki_alert_rules/kratos_high_severity_log.rule
+++ b/src/loki_alert_rules/kratos_high_severity_log.rule
@@ -2,37 +2,37 @@ groups:
 - name: KratosHighSeverityLog
   rules:
   - alert: ErrorLevelLog
-    expr: sum(count_over_time({%%juju_topology%%} |= `"severity":"error"` [5m])) < 100 and sum(count_over_time({%%juju_topology%%} |= `"severity":"error"` [5m])) > 0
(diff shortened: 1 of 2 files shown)


shipperizer (commented):
see comments in https://github.com/canonical/hydra-operator/pull/92 as they apply here too

shipperizer (approved):
same as for https://github.com/canonical/hydra-operator/pull/92#pullrequestreview-1547466966

content looks good

PR #109, opened 2023-08-01
Use configMap to store config

nsklikas:
IAM-362:
- Start using configMap for storing the kratos config
- Fix migration logic


shipperizer (approved):

src/charm.py
```diff
@@ -330,24 +340,10 @@ def _render_conf_file(self) -> str:
...
-
-    def _push_default_files(self) -> None:
-        for schema_file in self._mappers_local_dir_path.iterdir():
-            self._push_file(self._mappers_dir_path / schema_file.name, src=schema_file)
-
-        for schema_file in self._identity_schemas_local_dir_path.iterdir():
-            self._push_file(
-                self._identity_schemas_default_dir_path / schema_file.name, src=schema_file
-            )
+    @backoff.on_exception(backoff.expo, ApiError, max_tries=5, logger=logger)
```
wood-push-melon, line 343: Have two questions about other CRUD operations on ConfigMaps:

- do we need to provide the same backoff retry logic to the other operations, like create, delete, etc.?
- what I can see is that we probably should put the retry logic into the `ConfigMapHandler`. We don't need to render the config file several times due to the k8s client API errors.

src/charm.py
```diff
@@ -383,35 +391,53 @@ def _get_juju_config_identity_schema_config(self) -> Optional[Tuple[str, Dict]]:
...
         if default_schema_id not in schemas:
             raise RuntimeError(f"Default schema `{default_schema_id}` can't be found")
+        schemas = {
+            schema_id: f"base64://{base64.b64encode(schema.encode()).decode()}"
+            for schema_id, schema in schemas.items()
+        }
         return default_schema_id, schemas
 
-    def _get_identity_schema_config(self) -> Optional[Tuple[str, Dict]]:
+    def _get_identity_schema_config(self) -> Tuple[str, Dict]:
```
wood-push-melon, line 440: Saw some common patterns in each identity schema source `get` function. Could be an overkill in the Python world, but I think it would be better when more sources of identity schema show up.

- Extract the identity schema config as a standalone base class. It has a `render` abstract method which can be overridden by different subclasses
- Create different classes that inherit the base class by overriding the `render` method, e.g. `DefaultIdentitySchemaConfig`
- Apply a chain of calling each class's `render` method with a specific order. The chain will return early if some of the handlers return an identity schema

We can control the chain order very easily and extend it easily with more subclasses.

Just a thought. No need to make any changes.

src/charm.py
```diff
@@ -467,6 +526,15 @@ def _create_secret(self) -> Optional[Secret]:
         juju_secret = self.model.app.add_secret(secret, label=SECRET_LABEL)
         return juju_secret
 
+    def _migration_is_needed(self) -> Optional[bool]:
```
wood-push-melon, line 529: Just have a minor question about why we return an `Optional`. We directly return `bool`. Unless the callers have different handling logic for `None` and `False`?

```
if not self._peers:
    return False
```

src/charm.py
```diff
@@ -561,57 +636,80 @@ def _update_kratos_endpoints_relation_data(self, event: RelationEvent) -> None:
...
+            ],
+            "volumes": [
+                {
+                    "name": "config",
+                    "configMap": {"name": self._kratos_config_map_name},
+                },
+            ],
+        }
+        patch = {"spec": {"template": {"spec": pod_spec_patch}}}
+        self.client.patch(StatefulSet, name=self.meta.name, namespace=self.model.name, obj=patch)
```
wood-push-melon, line 664: Have a question about the `client` here. I saw this `client` is also used in the `ConfigMapHandler` class.

I think it would be better to isolate the client of a third-party library into an individual client class. In this way, the changes in the third-party APIs will not infiltrate different places, causing changes in multiple places.

src/config_map.py
```diff
@@ -0,0 +1,128 @@
...
+
+from lightkube import ApiError, Client
+from lightkube.models.meta_v1 import ObjectMeta
+from lightkube.resources.core_v1 import ConfigMap
+from ops.charm import CharmBase
+
+logger = logging.getLogger(__name__)
+
+
+class ConfigMapHandler(CharmBase):
```
wood-push-melon, line 17: Sorry, I am confused about why we inherited the `CharmBase` class.

I guess it would be better if we layer in the following way:

|     kubeclient    |    configmap manager    |  charm    |

- kubeclient will be responsible for interacting with the k8s cluster using the `client` in the lightkube library. It will provide other callers with stable APIs, like `delete_config_map`, `patch_statefulsets`, etc. This will isolate the potential API changes from the third-party, and also we can easily swap the k8s client library in the future.
- configmap manager will be using the kubeclient APIs to compose biz logic layer, e.g. `create_all`, `delete_all`. 
- we extract the config map as either a protocol or base class with `create`, `update`, `delete` methods, and create subclasses inheriting the config map, e.g. `KratosConfigMap`, `ProviderConfigMap` to override the methods
- We can register all the types of the config map in the configmap manager, and then loop through the registry to create/delete them all in the `create/delete_all` methods. This way we can easily control the registry without changing the configmap manager's methods.
- This will give us scalability if more types of config maps show up in the future without changing configmap manager.

I think we could improve this later in other PRs.

src/kratos.py
```diff
@@ -137,6 +138,25 @@ def run_migration(self, timeout: float = 120) -> str:
         ]
         return self._run_cmd(cmd, timeout=timeout)
 
+    def get_version(self) -> Dict[str, str]:
+        """Get the version of the hydra binary."""
```
wood-push-melon, line 142: typo: `hydra` -> `kratos`

src/kratos.py
```diff
@@ -137,6 +138,25 @@ def run_migration(self, timeout: float = 120) -> str:
...
+        stdout = self._run_cmd(cmd)
+
+        # Output has the format:
+        # Version:    {version}
+        # Build Commit:   {hash}
+        # Build Timestamp: {time}
+        out_re = r"Version:[ ]*(.+)\nBuild Commit:[ ]*(.+)\nBuild Timestamp:[ ]*(.+)"
+        versions = re.findall(out_re, stdout)[0]
+
+        return {
```
wood-push-melon, line 154: Nitpick:

I think we can just return what the caller needs: the `version`. The callers don't care about other information and they don't necessarily need to know how to obtain it (e.g. do a dict lookup).

src/charm.py
nsklikas: The main reason for the backoff is that the API call may fail because someone else updated the configMap while we were creating the config. In that case, we need to re-generate the config to take into account anything that was changed.

src/charm.py
nsklikas: The callers don't have different logic, because they check if the peer relation exists themselves. I wanted to be able distinguish between whether the migration has not run or if the peer relation is not ready, but this is not used right now. I could remove this check altogether.

src/charm.py
```diff
@@ -503,6 +503,31 @@ def _run_sql_migration(self) -> bool:
...
+        peers.data[self.app][key] = json.dumps(data)
+
+    def _get_peer_data(self, key: str) -> Union[Dict, str]:
+        """Retrieve information from the peer data bucket."""
+        if not (peers := self._peers):
+            return {}
+        data = peers.data[self.app].get(key, "")
+        return json.loads(data) if data else {}
+
+    def _pop_peer_data(self, key: str) -> Dict:
```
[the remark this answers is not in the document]
nsklikas: No, it was copied over from hydra, but it should be used to pop the migration version when the database is deleted (I will add this)

src/kratos.py
nsklikas: Fair point

src/charm.py
nsklikas: It makes sense what you are suggesting, I don't think we will need to add more sources for this in the future, but it would be nice to just separate this logic from the rest of the charm code. If it's OK with you though I would like to do that in a later PR.

src/config_map.py
nsklikas: The `CharmBase` was left over, I removed it.

I tried to implement this abstraction except for the adapter around the lightkube client. We can work on that on a later PR if that's OK with you.

src/charm.py
wood-push-melon: yeah, definitely don't change further in this PR. We can improve it later.

src/charm.py
wood-push-melon: ic~ understood. I would like to say try not to use two falsy return values `None` and `False` together. The caller might be confused by those two return values. Maybe we could do the following:

```
def is_peer_relation_ready(self):
    return self._peers is not None

def is_migration_needed(self):
    return self._get_peer_data(PEER_KEY_DB_MIGRATE_VERSION) != self.kratos.get_version()

# In the caller function we do:
if self.is_peer_relation_ready() and self.is_migration_needed():
    do_something()
```

src/charm.py
```diff
@@ -339,44 +357,35 @@ def _render_conf_file(self) -> str:
...
-
-    def _push_default_files(self) -> None:
-        for schema_file in self._mappers_local_dir_path.iterdir():
-            self._push_file(self._mappers_dir_path / schema_file.name, src=schema_file)
-
-        for schema_file in self._identity_schemas_local_dir_path.iterdir():
-            self._push_file(
-                self._identity_schemas_default_dir_path / schema_file.name, src=schema_file
-            )
+    @retry(
```
wood-push-melon, line 380: Have a question: is it possible that we can remove the retry mechanism since we move the `dsn` to env vars?

src/charm.py
```diff
@@ -607,6 +668,8 @@ def _on_leader_elected(self, event: LeaderElectedEvent) -> None:
 
     def _on_pebble_ready(self, event: PebbleReadyEvent) -> None:
         """Event Handler for pebble ready event."""
+        self._patch_statefulset()
```
wood-push-melon, line 671: Just have a small question: any reason we patch the StatefulSet in the `x-pebble-ready` event handle?

wood-push-melon:
Generally LGTM. Just have two small questions.

wood-push-melon (dismissed):

src/charm.py
nsklikas: The reason for the retry is that if someone (admin UI) else tries to patch the configmap at the same time, we will get an error (because we will have an older revision). If this happens, we need to either catch the error and defer the event or retry. 

Since it is quite possible that our request gets a transient error (as described above), I think that it makes sense to retry instead of failing right away.

src/charm.py
```diff
@@ -339,44 +357,35 @@ def _render_conf_file(self) -> str:
...
-            )
+    @retry(
+        wait=wait_exponential(multiplier=3, min=1, max=10),
+        stop=stop_after_attempt(5),
+        reraise=True,
+        before=before_log(logger, logging.DEBUG),
+    )
+    def _update_config(self) -> None:
+        conf = self._render_conf_file()
+        self.kratos_configmap.update({"kratos.yaml": conf})
```
shipperizer, line 389: do we need the extra assignment?
performance wise I don't think it makes a big difference in fairness so happy to keep it

src/charm.py
nsklikas: AFAIK it is common practice to do it in that event.

I think that the reason the `pebble-ready` event is considered ideal for this is that we want all our charms to patch the statefulSet on the same event, so it should not be a "custom" event (like database created). Also we want it to happen early, as it causes the pod to restart. IMO the only other event where it makes sense to patch the statefulSe is `install`. According to the [juju docs](https://juju.is/docs/sdk/install-event#heading--emission-sequence), the `install` event seems like a fine place to patch the statefulSet as well, BUT it is not called when the charm is upgraded so we would have to patch the statefulSet on both `install` and `upgrade-charm`. Tbh I have no problem with that but I decided to follow what (I think) is common practice.

shipperizer (approved):

src/charm.py
nsklikas: I just didn't want to have too much happening in a single line, no problem with changing it if you think it would be cleaner that way.

wood-push-melon (approved):

PR #110, opened 2023-08-01
doc: provide the data migration guides

wood-push-melon:
This pull request provides the SQL migration guides for Kratos Charm.


wood-push-melon:
Need to transfer it to Google Doc since this documentation should be managed in discourse and published in Charmhub.

PR #123, opened 2023-08-31
IAM-460-Update login-ui relation

bencekov:
As part of IAM-460, I took out the unused pages from the login-ui backend as well as from the login_ui_endpoints integration lib. This pr updates the charm according to the changes in login_ui_endpoints.py.


templates/kratos.yaml.j2
```diff
@@ -27,17 +27,6 @@ selfservice:
...
         {%- endif %}
-        {%- if registration_ui_url %}
-        registration:
-            enabled: True
-            ui_url: {{ registration_ui_url }}
-            {%- if oidc_providers %}
-            after:
-                oidc:
-                    hooks:
-                    - hook: session
```
nsklikas, line 38: When user logs in for the first time, they are registered. If the session hook is not enabled, they will not be logged in (see https://www.ory.sh/docs/kratos/hooks/configure-hooks#available-actions). 

We can't remove this

nsklikas (dismissed):

src/charm.py
```diff
@@ -325,7 +325,7 @@ def _render_conf_file(self) -> str:
             cookie_secrets=[self._get_secret()],
             log_level=self._log_level,
             mappers_path=str(self._mappers_dir_path),
-            default_browser_return_url=self._get_login_ui_endpoint_info("default_url"),
+            default_browser_return_url=self._get_login_ui_endpoint_info("login_url"),
```
nsklikas, line 328: We should either use the `default_url` here or remove it from the databag (see https://github.com/canonical/identity-platform-login-ui-operator/pull/60/files#r1314575045)

bencekov: done

nsklikas (approved):

PR #158, opened 2023-09-15
Fix invalid config when no OIDC Provider

nsklikas:
IAM-491

### DIFF
--- a/templates/kratos.yaml.j2
+++ b/templates/kratos.yaml.j2
@@ -17,7 +17,7 @@ selfservice:
     {%- endif %}
     default_browser_return_url:
         {{ default_browser_return_url | d("http://example-default-return-url.com", true) }}
-    {%- if error_ui_url or login_ui_url or registration_ui_url %}
+    {%- if error_ui_url or login_ui_url or oidc_providers %}
     flows:
         {%- if error_ui_url %}
         error:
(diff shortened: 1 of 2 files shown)


templates/kratos.yaml.j2
```diff
@@ -27,13 +27,13 @@ selfservice:
...
+        {%- if oidc_providers %}
         registration:
-            {%- if oidc_providers %}
             after:
                 oidc:
                     hooks:
                     - hook: session
-            {%- endif %}
+        {%- endif %}
     {%- endif %}
```
natalian98, line 37: This block depends on `{%- if error_ui_url or login_ui_url or registration_ui_url %}`, but there is no `registration_ui_url` property in the kratos charm. Should we change it to `oidc_providers`?
```yaml
{%- if error_ui_url or login_ui_url or oidc_providers %}
flows:
    {%- if error_ui_url %}
    error:
        ui_url: {{ error_ui_url }}
    {%- endif %}
    {%- if login_ui_url %}
    login:
        ui_url: {{ login_ui_url }}
    {%- endif %}
    {%- if oidc_providers %}
    registration:
        after:
            oidc:
                hooks:
                - hook: session
    {%- endif %}
{%- endif %}
```

nsklikas: Yes, good catch.

natalian98 (approved):

PR #161, opened 2023-09-22
Improve migration logic

nsklikas:
Improvements around the migration logic. The goal is to bring the Kratos charm closer to the logic in the Hydra charm.

Changes:
- JSON dump the contents in the peer relation, this will make backwards compatibility much easier to achieve. Currently the charm stores the migration version in the peer relation as a string: `{"db_migrate_version": "0.11.1"}`, after these changes the version will be json dumped: `{"db_migrate_version": '"0.11.1"'}`. This is a breaking change and will cause the charm to go to an error state if you try to upgrade from the previous version to this.
- Use DSN from env var to run migration instead of having to read the config file. This will make it much simpler to run the migrations once the configMap is introduced (we won't have to wait for the configMap to update in order to run the migrations).
- Do not start the service in the database_created hook, instead call `_handle_status_update_config` to start the service. This introduces a single function that will be responsible for starting the Kratos service. This is a small change, whose only purpose is to simplify the code and remove code duplication.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -13,7 +13,7 @@ from functools import cached_property
 from os.path import join
 from pathlib import Path
 from secrets import token_hex
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 
 import requests
 from charms.data_platform_libs.v0.data_interfaces import (
@@ -63,6 +63,7 @@ from ops.model import (
     BlockedStatus,
     MaintenanceStatus,
     ModelError,
+    Relation,
     Secret,
     SecretNotFoundError,
     WaitingStatus,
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -580,58 +636,46 @@ def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
...
+            self._handle_status_update_config(event)
+            return
 
-        peer_relation = self.model.relations[PEER_RELATION_NAME]
-        if not peer_relation:
-            event.defer()
-            self.unit.status = WaitingStatus("Waiting for peer relation.")
+        if not self.unit.is_leader():
+            logger.info("Unit does not have leadership")
+            self.unit.status = WaitingStatus("Unit waiting for leadership to run the migration")
```
natalian98, line 659: Should we also check if the migration needs to be run on leader-elected event?

shipperizer (approved):

PR #164, opened 2023-09-25
Expose app version to juju

nsklikas:
IAM-495


nsklikas:
This change will expose the kratos version as seen below:
image

Most of our charmed applications do not have the `v` prefix, should we filter that out?

shipperizer (approved):

shipperizer (approved):

shipperizer (approved):

PR #169, opened 2023-10-12
Add public ingress url to allowed return urls

bencekov:
When the public ingress is available, its url is added to the allowed return urls in Kratos' config file.

If the url of the public ingress is http://ingress:80/path/subpath it gets added to the config file as https://ingress/*.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -14,6 +14,7 @@ from os.path import join
 from pathlib import Path
 from secrets import token_hex
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
+from urllib.parse import urlparse
 
 import requests
 from charms.data_platform_libs.v0.data_interfaces import (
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -340,12 +340,19 @@ def _render_conf_file(self) -> str:
 
         default_schema_id, schemas = self._get_identity_schema_config()
         login_ui_url = self._get_login_ui_endpoint_info("login_url")
+
+        allowed_return_urls = []
+        if self._public_url:
+            allowed_return_urls.append(domain_url(self._public_url))
+        if login_ui_url:
+            allowed_return_urls.append(login_ui_url)
```
nsklikas, line 348: I don't think this is needed anymore (the login UI is in the same domain)

bencekov: true, thanks for the catch!

src/charm.py
```diff
@@ -361,12 +361,18 @@ def _render_conf_file(self) -> str:
         login_ui_url = self._get_login_ui_endpoint_info("login_url")
         mappers = self._get_claims_mappers()
         cookie_secrets = self._get_secret()
+
+        allowed_return_urls = []
+        if self._public_url:
+            allowed_return_urls.append(domain_url(self._public_url))
```
shipperizer, line 367:
```suggestion
            allowed_return_urls.append(urlparse(self.public_url)._replace(path="*", params="", query="", fragment="").geturl())
```


I would put it inline, jumping across module for a function that is a little wrapper of a built-in function doesn't seem that intuitive

src/charm.py
```diff
@@ -361,12 +362,22 @@ def _render_conf_file(self) -> str:
...
+
         rendered = template.render(
             cookie_secrets=[cookie_secrets] if cookie_secrets else None,
             log_level=self._log_level,
             mappers=mappers,
-            default_browser_return_url=self._get_login_ui_endpoint_info("login_url"),
-            allowed_return_urls=[login_ui_url] if login_ui_url else [],
+            mappers_path=str(self._mappers_dir_path),
+            default_browser_return_url=login_ui_url,
+            allowed_return_urls=allowed_return_urls,
```
nsklikas, line 380:
```suggestion
            default_browser_return_url=login_ui_url,
            allowed_return_urls=allowed_return_urls,
```

shipperizer:
there are 4 commits and 16 additions, which are not really distinguishable/easy to separate

I'd say make one commit out of it

src/charm.py
bencekov: done

nsklikas (approved):

PR #179, opened 2024-02-07
Add `kratos-info` interface

natalian98:
This PR adds a new `kratos-info` interface providing endpoints and configmaps details in a single relation.
It deprecates the `kratos-endpoint-info` interface which has a narrow scope.

### DIFF
--- a/lib/charms/harness_extensions/v0/capture_events.py
+++ b/lib/charms/harness_extensions/v0/capture_events.py
@@ -1,0 +1,86 @@
+'''This is a library providing a utility for unittesting events fired on a
+Harness-ed Charm.
+
+Example usage:
+
+>>> from charms.harness_extensions.v0.capture_events import capture
+>>> with capture(RelationEvent) as captured:
+>>>     harness.add_relation('foo', 'remote')
+>>> assert captured.event.unit.name == 'remote'
+'''
+
+# The unique Charmhub library identifier, never change it
+LIBID = "9fcdab70e26d4eee9797c0e542ab397a"
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 3
+
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+from contextlib import contextmanager
+from typing import Generic, Iterator, Optional, Type, TypeVar
+
+from ops.charm import CharmBase
+from ops.framework import EventBase
+
+_T = TypeVar("_T", bound=EventBase)
+
+
+@contextmanager
+def capture_events(charm: CharmBase, *types: Type[EventBase]):
+    """Capture all events of type `*types` (using instance checks)."""
+    allowed_types = types or (EventBase,)
+
+    captured = []
+    _real_emit = charm.framework._emit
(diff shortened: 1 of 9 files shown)


lib/charms/kratos/v0/kratos_info.py
nsklikas, line 1: Since we are creating a new library, I think it is worth considering whether we should try to spend some time to design the databag format (for example per @wood-push-melon spec)

src/charm.py
```diff
@@ -709,22 +729,26 @@ def _get_tracing_endpoint_info(self) -> str:
...
-            self._admin_url
-            or f"http://{self.app.name}.{self.model.name}.svc.cluster.local:{KRATOS_ADMIN_PORT}"
-        )
-        public_endpoint = (
-            self._public_url
-            or f"http://{self.app.name}.{self.model.name}.svc.cluster.local:{KRATOS_PUBLIC_PORT}"
-        )
+        (admin_endpoint, public_endpoint) = self._kratos_endpoints
+        providers_cofigmap_name = "providers"
+        schemas_cofigmap_name = "identity-schemas"
```
nsklikas, line 737: Let's not hardcode this
```suggestion
        providers_configmap_name = self.providers_configmap.name
        schemas_configmap_name = self.schemas_configmap.name
```

src/charm.py
```diff
@@ -709,22 +729,26 @@ def _get_tracing_endpoint_info(self) -> str:
...
-            public_endpoint.replace("https", "http"),
+        self.info_provider.send_info_relation_data(
+            admin_endpoint,
+            public_endpoint,
+            providers_cofigmap_name,
+            schemas_cofigmap_name,
+            configmaps_namespace,
         )
+
+    def _update_kratos_endpoints_relation_data(self, event: RelationEvent) -> None:
```
nsklikas, line 748: Let's add a `TODO` to remove this (btw when are we going to remove this?)

nsklikas (dismissed):

nsklikas (dismissed):

src/charm.py
natalian98: We should remove this once we switch to using `kratos-info` in login ui.
Comment added

lib/charms/harness_extensions/v0/capture_events.py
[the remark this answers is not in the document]
natalian98: Perhaps it would make sense to remove `capture_events` from lib and just move it to tests directory?

lib/charms/kratos/v0/kratos_info.py
```diff
@@ -0,0 +1,169 @@
...
+            relation.data[self._charm.app].update(info_databag)
+
+
+class KratosInfoRelationError(Exception):
+    """Base class for the relation exceptions."""
+    pass
+
+
+class KratosInfoRelationMissingError(KratosInfoRelationError):
+    """Raised when the relation is missing."""
```
BarcoMasile, line 124: suggestion: how about we remove the word `Kratos` from errors so it's less redundant?

BarcoMasile (dismissed):

lib/charms/kratos/v0/kratos_info.py
natalian98: I would prefer to keep it to avoid confusions with other similar interfaces

BarcoMasile (approved):

PR #183, opened 2024-04-18
Update kratos_external_idp_integrator lib

nsklikas:
Update kratos_external_idp_integrator lib and properly handle provider removal.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -37,6 +37,7 @@ from charms.kratos.v0.kratos_endpoints import KratosEndpointsProvider
 from charms.kratos.v0.kratos_info import KratosInfoProvider
 from charms.kratos_external_idp_integrator.v0.kratos_external_provider import (
     ClientConfigChangedEvent,
+    ClientConfigRemovedEvent,
     ExternalIdpRequirer,
 )
 from charms.loki_k8s.v0.loki_push_api import LogProxyConsumer, PromtailDigestError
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -881,17 +885,22 @@ def _on_client_config_changed(self, event: ClientConfigChangedEvent) -> None:
...
+
+        self._handle_status_update_config(event)
+        self.external_provider.set_relation_registered_provider(
+            join(public_url, f"self-service/methods/oidc/callback/{event.provider_id}"),
+            event.provider_id,
+            event.relation_id,
+        )
+
+    def _on_client_config_removed(self, event: ClientConfigRemovedEvent) -> None:
+        self.unit.status = MaintenanceStatus(f"Removing external provider")
```
natalian98, line 901: f string is not needed here

tests/unit/test_charm.py
```diff
@@ -818,6 +820,138 @@ def test_on_config_changed_with_ingress(
...
+) -> None:
+    setup_peer_relation(harness)
+    setup_postgres_relation(harness)
+    setup_ingress_relation(harness, "public")
+    (_, login_databag) = setup_login_ui_relation(harness)
+
+    relation_id, data = setup_external_provider_relation(harness)
+    harness.remove_relation(relation_id)
+    container = harness.model.unit.get_container(CONTAINER_NAME)
+    harness.charm.on.leader_elected.emit()
```
natalian98, line 837: no need to call this event as harness triggers it anyway when the leader is set (see https://github.com/canonical/kratos-operator/blob/55ba14cdb9d71220445a6ba037e16c0bef3790f8/tests/unit/conftest.py#L23)

nsklikas: There are some tests that set the leader to `False` (eg https://github.com/canonical/kratos-operator/blob/main/tests/unit/test_charm.py#L572) and the state persists. If I remove this the test fails.

This probably means that we should refactor our tests, but this is not the right time

PR #193, opened 2024-06-27
IAM-942: internal ingress support

shipperizer:
IAM-942: introduce internal-ingress via `traefik-route`

- **feat: introduce internal ingress**
- **fix: use internal ingress if set, otherwise stick with k8s networking**

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 # TODO: Remove when pr https://github.com/go-macaroon-bakery/py-macaroon-bakery/pull/95 gets merged
 macaroonbakery==1.3.2
 cosl==0.0.7
-ops==2.10.0
+ops==2.14.0
 lightkube===0.15.0
 lightkube-models==1.29.0.6
 Jinja2==3.1.3
--- a/src/charm.py
+++ b/src/charm.py
@@ -49,6 +49,7 @@ from charms.traefik_k8s.v2.ingress import (
     IngressPerAppRequirer,
     IngressPerAppRevokedEvent,
 )
+from charms.traefik_route_k8s.v0.traefik_route import TraefikRouteRequirer
 from jinja2 import Template
 from lightkube import Client
 from lightkube.resources.apps_v1 import StatefulSet
@@ -62,6 +63,7 @@ from ops.charm import (
     PebbleReadyEvent,
     RelationDepartedEvent,
     RelationEvent,
+    RelationJoinedEvent,
     RemoveEvent,
     UpgradeCharmEvent,
 )
@@ -81,6 +83,7 @@ from tenacity import before_log, retry, stop_after_attempt, wait_exponential
 
 import config_map
 from config_map import IdentitySchemaConfigMap, KratosConfigMap, ProvidersConfigMap
+from constants import INTERNAL_INGRESS_RELATION_NAME
 from kratos import KratosAPI
 from utils import dict_to_action_output, normalise_url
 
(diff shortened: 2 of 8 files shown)


requirements.txt
```diff
@@ -1,7 +1,7 @@
 # TODO: Remove when pr https://github.com/go-macaroon-bakery/py-macaroon-bakery/pull/95 gets merged
 macaroonbakery==1.3.2
```
BarcoMasile, line 2: out-of-scope: Apparently the PR mentioned in the TODO is merged. Should we try to remove this dependency?

src/utils.py
```diff
@@ -51,6 +51,9 @@ def normalise_url(url: str) -> str:
     """
     parsed_url = urlparse(url)
 
+    # latest versions of traefik automatically redirect to https is ceritficate relation is
```
BarcoMasile, line 54:
```suggestion
    # latest versions of traefik automatically redirect to https if ceritficate relation is
```

requirements.txt
shipperizer: let's make an issue

shipperizer: #195

src/charm.py
```diff
@@ -343,21 +366,104 @@ def _admin_url(self) -> Optional[str]:
...
+                }
+            },
+        }
+
+        return {"http": {"routers": routers, "services": services, "middlewares": middlewares}}
+
     @property
     def _kratos_endpoints(self) -> Tuple[str, str]:
         admin_endpoint = (
-            self._admin_url
```
natalian98, line 349: we can remove `_admin_url` property since it's no longer used

src/charm.py
```diff
@@ -343,21 +366,104 @@ def _admin_url(self) -> Optional[str]:
...
+        external_path = f"{self.model.name}-{self.model.app.name}"
+
+        middlewares = {
+            f"juju-sidecar-noprefix-{self.model.name}-{self.model.app.name}": {
+                "stripPrefix": {"forceSlash": False, "prefixes": [f"/{external_path}"]},
+            },
+        }
+
+        routers = {
+            "juju-{}-{}-admin-api-router".format(self.model.name, self.model.app.name): {
```
natalian98, line 391: nitpick: can we use fstrings instead of format?

natalian98 (commented):
If I understand correctly, this PR deprecates the `admin_ingress` relation in favor of `internal_ingress`. 
Should we remove the old relation then?

shipperizer:
> If I understand correctly, this PR deprecates the `admin_ingress` relation in favor of `internal_ingress`. Should we remove the old relation then?

yes, we will, just would say we should do it in a separate pr and trigger the new track from there

src/charm.py
```diff
@@ -251,6 +263,21 @@ def __init__(self, *args: Any) -> None:
         self.framework.observe(self.tracing.on.endpoint_changed, self._on_config_changed)
         self.framework.observe(self.tracing.on.endpoint_removed, self._on_config_changed)
 
+        self.framework.observe(
+            self.on[INTERNAL_INGRESS_RELATION_NAME].relation_joined,
+            self._configure_internal_ingress,
+        )
```
natalian98, line 269: no need to observe `relation-joined`, `relation-changed` is emitted right after it (see https://juju.is/docs/sdk/relation-name-relation-joined-event)

shipperizer: it seems to be breaking the tests, not sure if it's because the harness is badly set or something else

will address this in a separate PR if it's worth fixing

natalian98 (approved):

PR #199, opened 2024-07-30
feat: add account management actions

natalian98:
This PR adds juju actions for sessions invalidation, password reset (with either recovery code or a password hashed with the default bcrypt algorithm) and mfa credentials reset.
It removes password reset with link method because it is [considered deprecated](https://www.ory.sh/docs/kratos/self-service/flows/account-recovery-password-reset#comparison).

Password will still be required for admin account creation because the credentials must be provided in `kratos import identities`.

### DIFF
--- a/src/kratos.py
+++ b/src/kratos.py
@@ -9,6 +9,7 @@ import re
 from os.path import join
 from typing import Dict, List, Optional
 
+import bcrypt
 import requests
 from ops.model import Container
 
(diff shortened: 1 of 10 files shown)


actions.yaml
```diff
@@ -38,8 +38,43 @@ create-admin-account:
     phone_number:
       description: The admin's phone number
       type: string
-  # TODO: Remove required password once password resetting is implemented
```
natalian98, line 41: This can't be done because according to [docs](https://www.ory.sh/docs/kratos/manage-identities/import-user-accounts-identities)
> when you import identities, you must provide the credentials field.

src/kratos.py
```diff
@@ -117,16 +118,62 @@ def recover_password_with_code(self, identity_id: str, expires_in: str = "1h") -
...
+        """Set identity's password to a provided value."""
+        identity = self.get_identity(identity_id)
+        traits = identity.get("traits")
+        schema_id = identity.get("schema_id")
+        state = identity.get("state")
+
+        credentials = {
+            "password": {
+                "config": {
+                    "hashed_password": bcrypt.hashpw(password.encode("utf-8"), bcrypt.gensalt()).decode("utf-8")
```
natalian98, line 131: Bcrypt is the [default](https://www.ory.sh/docs/self-hosted/kratos/configuration/password#bcrypt) algorithm for hashing, let me know it should be changed

actions.yaml
BarcoMasile: Let's add this in the PR description!

BarcoMasile (approved):
Minor comments, LGTM

actions.yaml
```diff
@@ -38,8 +38,43 @@ create-admin-account:
...
+  description: |
+    Reset identity's second authentication factor using either the identity ID or user email.
+  params:
+    identity-id:
+      description: The Identity ID
+      type: string
+    email:
+      description: The user's email
+      type: string
+    mfa-type:
```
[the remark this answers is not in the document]
natalian98: I don't set a default on purpose, to avoid accidental deletion

nsklikas (commented):
Can we add a couple of integration tests for the new actions or would that require deploying the login ui?

nsklikas (approved):

PR #222, opened 2024-08-08
feat: support local identity provider

natalian98:
This PR adds support for local identity provider:
- updates identity schemas with email identifiers for the new flow methods
- adds config options to enable Kratos IdP and passwordless login method
- updates the config template
- bumps `ui-endpoint-info` lib
## Testing
1. Build kratos with `charmcraft pack`.
2. Render and deploy the identity platform bundle, replacing kratos with the locally built charm and bumping login ui revision (105 or higher).
`juju deploy ./bundle-edge.yaml --trust`
3. Deploy grafana with required relations:
```
juju deploy grafana-k8s
juju integrate grafana-k8s:ingress traefik-public
juju integrate grafana-k8s:oauth hydra
```
4. Create a user in kratos:
`juju run kratos/0 create-admin-account email=test@example.com password=test username=admin`
5. Go to https://{public-traefik-ip}/{model}-grafana-k8s to test the login flow.

To test the recovery code flow, run the `reset-password` action and go to the returned url. Self-service recovery flow doesn't work atm because there is no smtp courier, so no emails are sent. This will be addressed in the next PR.

To test MFA, log in or recover the account and go to https://{public-traefik-ip}/{model}-identity-platform-login-ui-operator/ui/setup_secure.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -103,6 +103,15 @@ config:
         Run Kratos on dev mode, it is needed if HTTPS is not set up. This should only be used for development purposes.
       type: boolean
       default: False
+    enable_local_idp:
+      description: Enable Kratos Identity Provider
+      type: boolean
+      default: True
+    enable_passwordless_login_method:
+      description: |
+        Enable passwordless authentication via webauthn. Requires `enable_local_idp=True`.
+      type: boolean
+      default: False
     identity_schemas:
       description: |
         A mapping of schema_id to identity schemas. For example:
--- a/templates/kratos.yaml.j2
+++ b/templates/kratos.yaml.j2
@@ -27,6 +27,21 @@ selfservice:
         login:
             ui_url: {{ login_ui_url }}
         {%- endif %}
+        {%- if settings_ui_url %}
+        settings:
+            ui_url: {{ settings_ui_url }}
+            required_aal: highest_available
+        {%- endif %}
+        {%- if recovery_ui_url %}
+        recovery:
+            enabled: True
+            ui_url: {{ recovery_ui_url }}
+            use: code
+            after:
+                default_browser_return_url: {{ default_browser_return_url }}
+                hooks:
+                    - hook: revoke_active_sessions
+        {%- endif %}
         {%- if oidc_providers %}
         registration:
             after:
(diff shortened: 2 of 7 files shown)


BarcoMasile (approved):

pik4ez-canonical (dismissed):
The PR was previously approved by Marco. Only a couple of minor dependency updates were [added](https://github.com/canonical/kratos-operator/compare/78e9c7bb1c7fa3de9334040d9c2dd366c42cb66c..81095626f2f1eb98a35a5685b100e32b3bb8e8eb) after that, so I'm approving the PR to remove the formal blocker.

Nikos [mentioned](https://matrix.to/#/!yqLAQacwXsgwHwgBvW:ubuntu.com/$9ZhJFBZkigV2UWw7EX_1Gcw1dP7RwGXvxD5oEyXHXrk?via=ubuntu.com) that he'll leave his feedback soon.

charmcraft.yaml
```diff
@@ -103,6 +103,15 @@ config:
         Run Kratos on dev mode, it is needed if HTTPS is not set up. This should only be used for development purposes.
       type: boolean
       default: False
+    enable_local_idp:
+      description: Enable Kratos Identity Provider
+      type: boolean
+      default: True
+    enable_passwordless_login_method:
+      description: |
+        Enable login with passwordless webauthn method. Requires `enable_local_idp=True`.
```
nsklikas, line 112: suggestion: would `webauthn` be a better word than `passwordless`? IMO passwordless is too generic, but I have no problem with keeping it as it is if you feel this is easier to understand

templates/kratos.yaml.j2
```diff
@@ -35,10 +50,38 @@ selfservice:
...
+                haveibeenpwned_enabled: False
+        totp:
+            enabled: True
+            config:
+                issuer: Identity Platform
+        {%- if enable_passwordless_login_method %}
+        webauthn:
+            enabled: True
+            config:
+                passwordless: True
```
nsklikas, line 76: Are there any use cases where we would want to use webauthn as MFA? Or we plan implement this in the login-ui?

Sorry I don't remember exactly what was agreed

nsklikas (approved):
LGTM overall, only a minor comment and a question

templates/kratos.yaml.j2
natalian98: We decided to only support webauthn as the first auth factor, as it can't be used as MFA simultaneously

charmcraft.yaml
nsklikas: I thought webauthn main use case is passwordless login, eg use your yubikey instead of having to remember a bunch of passwords. `passwordless` is probably more user friendly for users that have little experience with identity.

nsklikas, line 112: I would rephrase this as:
```
Enable passwordless authentication via webauthn. Requires `enable_local_idp=True`.
```

nsklikas (dismissed):

pik4ez-canonical (approved):

PR #283, opened 2024-10-14
fix: use [external] ingress on login endpoint field of kratos info

### LINKED ISSUE #267: Send public ingress url in `kratos-info` relation
### Enhancement Proposal

We should send the public ingress url in the `kratos-info` relation instead of the internal address:
https://github.com/canonical/kratos-operator/blob/2d46f228e572f4fa9779db2ece99f587288ca0cb/src/charm.py#L474-L477

This breaks the oathkeeper [integration](https://github.com/canonical/oathkeeper-operator/blob/2fc34750a984be66eb8b8ea96145ff33754e8a58/templates/oathkeeper.yaml.j2#L21)

shipperizer:
IAM-1066: addressing redirection issue for oathkeeper charm

closes #267

### DIFF
--- a/lib/charms/kratos/v0/kratos_info.py
+++ b/lib/charms/kratos/v0/kratos_info.py
@@ -40,6 +40,7 @@ Class SomeCharm(CharmBase):
 """
 
 import logging
+from os.path import join
 from typing import Dict, Optional
 
 from ops.charm import CharmBase, RelationCreatedEvent
@@ -53,7 +54,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 2
+LIBPATCH = 3
 
 RELATION_NAME = "kratos-info"
 INTERFACE_NAME = "kratos_info"
@@ -91,6 +92,7 @@ class KratosInfoProvider(Object):
         self,
         admin_endpoint: str,
         public_endpoint: str,
+        external_url: str,
         providers_configmap_name: str,
         schemas_configmap_name: str,
         configmaps_namespace: str,
@@ -100,11 +102,13 @@ class KratosInfoProvider(Object):
         if not self._charm.unit.is_leader():
             return
 
+        external_url = external_url if external_url.endswith("/") else external_url + "/"
+
         relations = self.model.relations[self._relation_name]
         info_databag = {
             "admin_endpoint": admin_endpoint,
             "public_endpoint": public_endpoint,
-            "login_browser_endpoint": f"{public_endpoint}/self-service/login/browser",
+            "login_browser_endpoint": join(external_url, "self-service/login/browser"),
             "sessions_endpoint": f"{public_endpoint}/sessions/whoami",
             "providers_configmap_name": providers_configmap_name,
             "schemas_configmap_name": schemas_configmap_name,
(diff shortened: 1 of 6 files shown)


lib/charms/kratos/v0/kratos_info.py
```diff
@@ -104,7 +106,7 @@ def send_info_relation_data(
         info_databag = {
             "admin_endpoint": admin_endpoint,
             "public_endpoint": public_endpoint,
-            "login_browser_endpoint": f"{public_endpoint}/self-service/login/browser",
+            "login_browser_endpoint": urljoin(external_url, "/self-service/login/browser"),
```
nsklikas, line 109: I am not sure if this will work, as usuall `external_url` includes a path:
```
>>> urljoin("https://traefik/model_charm", "/self-service/login/browser"),
('https://traefik/self-service/login/browser',)
```

IMO we should use `os.path.join` or parse the `external_url`

nsklikas (approved):

nsklikas (dismissed):

src/charm.py
```diff
@@ -931,9 +931,19 @@ def _update_kratos_info_relation_data(self, event: RelationEvent) -> None:
         configmaps_namespace = self.model.name
         mfa_enabled = self.config.get("enforce_mfa")
 
+        public_url = self._public_url
+        if public_url is None:
+            self.unit.status = BlockedStatus(
+                "Cannot add external provider without an external hostname. Please "
```
natalian98, line 937: It's just a nitpick, but the status message is a bit misleading. Adding an "internal" idp would also fail without the external url. Perhaps we could change it to sth like:
```suggestion
                "Cannot send integration data without an external hostname. Please "
```

natalian98 (approved):

nsklikas (approved):
lgtm. Only one nit-pick, I think that you don't need to enable metallb to run the integration tests, configuring traefik with an external_hostname should be enough (eg see [hydra](https://github.com/canonical/hydra-operator/blob/main/tests/integration/test_charm.py#L62))

src/charm.py
```diff
@@ -851,6 +852,14 @@ def _handle_status_update_config(self, event: HookEvent) -> None:
             event.defer()
             return
 
+        if self.model.relations[KRATOS_INFO_RELATION_NAME] and self._public_url is None:
+            self.unit.status = BlockedStatus(
+                "Cannot send integration data without an external hostname. Please "
+                "provide an ingress relation or an external_url config."
+            )
+            event.defer()
+            return
```
nsklikas, line 861: nit-pick: Maybe with this, you can undo the `raise_for_blocked` change in the tests.

```suggestion
        if self.model.relations[KRATOS_INFO_RELATION_NAME] and self.public_ingress.relation is None:
            self.unit.status = BlockedStatus(
                "Cannot send integration data without an external hostname. Please "
                "provide an ingress relation."
            )
            return
        elif self.model.relations[KRATOS_INFO_RELATION_NAME] and self._public_url is None:
            self.unit.status = WaitingStatus(
                "Waiting for ingress"
            )
            event.defer()
            return
```

src/constants.py
```diff
@@ -31,6 +33,7 @@
 DB_RELATION_NAME = "pg-database"
 HYDRA_RELATION_NAME = "hydra-endpoint-info"
 LOGIN_UI_RELATION_NAME = "ui-endpoint-info"
+KRATOS_INFO_RELATION_NAME = RELATION_NAME
```
nsklikas, line 36: Let's hard code the relation name here to keep it consistent. This name comes from charmcraft.yaml, not the relation lib

nsklikas (approved):
couple of nit-picks

nsklikas (approved):

PR #304, opened 2024-12-02
feat: replace traefik with istio-ingress for public ingress

wood-push-melon:


src/constants.py
```diff
@@ -24,7 +24,7 @@
 
 # Integration constants
 PEER_RELATION_NAME = "kratos-peers"
-INTERNAL_INGRESS_RELATION_NAME = "internal-ingress"
+PUBLIC_INGRESS_INTEGRATION_NAME = "public-ingress"
```
pik4ez-canonical, line 27: As far as we removed the internal ingress altogether, does it make sense to name the remaining one just "ingress", and not "public_ingress"?

pik4ez-canonical (approved):
A minor suggestion, otherwise LGTM.

src/constants.py
BarcoMasile: +1

wood-push-melon: For now I think keep it because there is another ingress relation called `admin-ingress` which typically is not often used at the current moment. And also we might add the internal ingress back if needed.

BarcoMasile (approved):

PR #326, opened 2025-01-29
Set a default return url for webauthn settings

natalian98:
This PR changes the default return url for webauthn method, so that the user remains on the setup page after removing or adding a key. This will not influence any redirect when setup is enforced during login.
fixes https://github.com/canonical/identity-platform-login-ui/issues/381


BarcoMasile (approved):

nsklikas (approved):
lgtm, but why do we need this?

natalian98:
I forgot to link the issue: https://github.com/canonical/identity-platform-login-ui/issues/381

PR #384, opened 2025-04-03
Update login_ui lib, add registration_url

nsklikas:
IAM-1418

The plan is that kratos will be fetching the ui endpoints from multiple charms. This is needed for the registration webhook service, because the user-verification charm will need to provide the registration_uri to kratos.

This will be tested as a whole in the user-verification-service-operator.


src/charm.py
```diff
@@ -590,6 +588,10 @@ def _smtp_connection_uri(self) -> str:
         else:
             return f"{method}://{username}:{password}@{server}:{port}/"
 
+    @property
+    def _login_ui_endpoints(self):
+        return
```
wood-push-melon, line 593: Should we remove this property?

nsklikas: good idea :P

wood-push-melon (dismissed):

shipperizer (approved):

PR #456, opened 2025-07-01
Store api token in juju secret

nsklikas:
IAM-1501


lib/charms/kratos/v0/kratos_registration_webhook.py
```diff
@@ -64,72 +59,49 @@ def deserialize_bool(v: str | bool) -> bool:
...
+    def auth_enabled(self) -> bool:
+        return all(
+            k
+            for k in [
+                self.auth_type,
+                self.auth_config_name,
+                self.auth_config_value_secret or self.auth_config_value,
+                self.auth_config_in,
+            ]
+        )
```
BarcoMasile, line 92: suggestion: feel free to ignore, this should work the same without the list comprehension, even with the `or`, I tried this in a python3 out of curiosity
```suggestion
        return all(
            [
                self.auth_type,
                self.auth_config_name,
                self.auth_config_value_secret or self.auth_config_value,
                self.auth_config_in,
            ]
        )
```

BarcoMasile (approved):

PR #463, opened 2025-07-04
Use config hash to restart service

nsklikas:
IAM-1612

There was an issue with the previous PR:
If multiple units of kratos are running, then one unit would change the configMap and the rest would not try to restart the service as the configmap will be up-to-date when they try to update it.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -630,6 +630,9 @@ class KratosCharm(CharmBase):
     def current_config_hash(self) -> Optional[int]:
         return self._stored.config_hash
 
+    def fetch_cm(self) -> str:
+        return self.kratos_configmap.get(load=False)["kratos.yaml"]
+
     @retry(
         wait=wait_exponential(multiplier=3, min=1, max=10),
         stop=stop_after_attempt(5),
--- a/src/config_map.py
+++ b/src/config_map.py
@@ -5,7 +5,7 @@
 
 import json
 import logging
-from typing import Dict
+from typing import Dict, Union
 
 from lightkube import ApiError, Client
 from lightkube.models.meta_v1 import ObjectMeta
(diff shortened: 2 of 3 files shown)


src/charm.py
```diff
@@ -615,15 +630,30 @@ def _smtp_connection_uri(self) -> str:
...
+    def _update_cm(self, content: str):
+        if self.unit.is_leader():
+            self.kratos_configmap.update({"kratos.yaml": content})
+
     def _update_config(self) -> None:
         conf = self._render_conf_file()
-        self._config_file_changed = self.kratos_configmap.update({"kratos.yaml": conf})
+        config_hash = utils.hash(conf)
+        if not config_hash != self.current_config_hash:
+            return
```
natalian98, line 651: feel free to ignore, but this seems clearer
```suggestion
        if config_hash == self.current_config_hash:
            return
```

src/charm.py
```diff
@@ -615,15 +630,30 @@ def _smtp_connection_uri(self) -> str:
...
+    def current_config_hash(self) -> Optional[int]:
+        return self._stored.config_hash
+
     @retry(
         wait=wait_exponential(multiplier=3, min=1, max=10),
         stop=stop_after_attempt(5),
         reraise=True,
         before=before_log(logger, logging.DEBUG),
     )
+    def _update_cm(self, content: str):
```
natalian98, line 643:
```suggestion
    def _update_cm(self, content: str) -> None:
```

src/config_map.py
```diff
@@ -75,22 +75,15 @@ def create(self) -> None:
         )
         self._client.create(cm)
 
-    def update(self, data: Dict, dump: bool = False) -> bool:
-        """Update the configMap.
-
-        Return value is a boolean indicating whether the cm data changed.
-        """
+    def update(self, data: Dict, dump: bool = False):
```
natalian98, line 78:
```suggestion
    def update(self, data: Dict, dump: bool = False) -> None:
```

natalian98:
q: would it be possible to add some unit test covering the changes?

shipperizer (approved):

PR #468, opened 2025-07-11
Use query param to fetch identity from email

### LINKED ISSUE #450: List identities by email does not support pagination
### Enhancement Proposal

List identities by email (probably) won't work if a lot of identities exist in Kratos. We should investigate the query params available in the Kratos list API, maybe we can avoid having to load all the identities in memory.

nsklikas:
IAM-1606

Closes #450


src/kratos.py
```diff
@@ -81,33 +82,20 @@ def delete_identity(self, identity_id: str) -> str:
...
-        return id_[0] if id_ else None
+        url = join(self.kratos_admin_url, "admin/identities")
+        r = requests.get(url, {"credentials_identifier": email})
+        r.raise_for_status()
+        identities = r.json()
+        if len(identities) > 1:
+            raise TooManyIdentitiesError()
+        if not identities:
+            return None
+        return r.json()[0]
```
BarcoMasile, line 98: Only nitpick I could find! Feel free to ignore as this adds zero value

```suggestion
        identities = r.json()
        if len(identities) > 1:
            raise TooManyIdentitiesError()
        if not identities:
            return None
        return identities[0]
```

BarcoMasile (approved):

PR #472, opened 2025-07-14
IAM-1572: use tf module in gh deployments

shipperizer:
- **feat: update juju_application resource name**
- **fix: use terraform module in deployment**
- **ci: update params for deployment wf**

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -79,6 +79,7 @@ jobs:
         revision: ${{ needs.revision.outputs.revision }}
         channel: ${{ needs.publish.outputs.channel}}
         application: kratos
+        tf-application-resource: module.application.juju_application.application
     secrets:
       CLIENT_ID: ${{ secrets.JIMM_DEV_CLIENT_ID }}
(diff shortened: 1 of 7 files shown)


.github/workflows/on_push.yaml
```diff
@@ -0,0 +1,19 @@
+name: On Push
+
+# On push to a "special" branch, we:
+# * always publish to charmhub at latest/edge/branchname
+# * always run tests
+# where a "special" branch is one of main/master or track/**, as
+# by convention these branches are the source for a corresponding
+# charmhub edge channel.
```
nsklikas, line 8: question: why is this workflow needed? Don't we already have a wf that does this on `main` and on `tracks`?

nsklikas (commented):
lgtm, but I have a question

.github/workflows/on_push.yaml
shipperizer: that should definitely not be there, pushed by accident

wood-push-melon (dismissed):
lgtm. just need to resolve the comments by Nikos.

PR #543, opened 2025-11-12
Stop service when database is gone

### LINKED ISSUE #534: Stop the service when database is gone
### Bug Description

When the database is gone, the service needs to be stopped. Otherwise the connection/session remains active

nsklikas:
IAM-1759

Closes #534

- Stop the service when the database is gone
- Introduce `collect_status` handler to simplify holistic handler function
- Introduce a fix for traefik_route, the bug was causing the charm to go into an error state when it was removed

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -16,12 +16,19 @@ from constants import (
 )
 
 
-@pytest.fixture(autouse=True)
-def mocked_k8s_resource_patch(mocker: MockerFixture) -> None:
-    mocker.patch(
+@pytest.fixture()
+def mocked_resource_patch(mocker: MockerFixture) -> MagicMock:
+    mocked = mocker.patch(
         "charms.observability_libs.v0.kubernetes_compute_resources_patch.ResourcePatcher",
         autospec=True,
     )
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -517,14 +493,67 @@ def _holistic_handler(self, event: EventBase) -> None:
 
         self.unit.status = ActiveStatus()
```
wood-push-melon: We need to remove this since we're going to use `on_collect_status`.

src/utils.py
```diff
@@ -69,3 +73,67 @@ def wrapped(charm: CharmBase) -> bool:
...
+    # TODO(nsklikas): Remove this when we update Kratos version and start using the passkey config to enable
+    # passkeys.
+    return not (
+        charm.charm_config["enable_oidc_webauthn_sequencing"]
+        and charm.charm_config["enable_passwordless_login_method"]
+    )
+
+
+# Condition failure causes early return without doing anything
+NOOP_CONDITIONS: tuple[Condition, ...] = (
```
wood-push-melon: A minor thing: just like the PR in hydra charm, we need to make type checker happy too in here.

PR #547, opened 2025-11-17
IAM-1785: use *-route relations

shipperizer:


wood-push-melon (approved):

wood-push-melon:
just a question: when should we release 2.0.0?

PR #558, opened 2025-11-21
Fix ca bundle generation

nsklikas:
- Fix bug where we would overwrite the ca bundle in the Kratos container with an empty file
- The bundle is re-generated only if something has changed

### DIFF
--- a/tests/unit/test_integrations.py
+++ b/tests/unit/test_integrations.py
@@ -30,6 +30,7 @@ from charms.smtp_integrator.v0.smtp import (
 )
 from charms.tempo_coordinator_k8s.v0.tracing import TracingEndpointRequirer
 from charms.traefik_k8s.v0.traefik_route import TraefikRouteRequirer
+from pytest_mock import MockerFixture
 from yarl import URL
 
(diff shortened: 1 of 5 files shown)


tests/unit/test_integrations.py
```diff
@@ -604,9 +605,70 @@ class TestTLSCertificates:
...
+        """Test that when the certificates have changed, subprocess.run is called."""
+        relation_certs = "ca_cert1\nca_cert2"
+        all_certs = relation_certs + "\nca_cert3"
+        mocked_local_bundle_file.read_text.return_value = "ca_cert1"
+        mocked_local_ca_bundle_file.read_text.return_value = all_certs
+        mocked_requirer.get_all_certificates.return_value = set(relation_certs.splitlines())
+
+        certificates = TLSCertificates.load(mocked_requirer)
+
+        mocked_subprocess.run.assert_called_once()
```
natalian98:
```suggestion
        mocked_subprocess.assert_called_once()
```

nsklikas: oops, fixed

BarcoMasile (approved):

natalian98 (approved):

PR #592, opened 2026-01-16
provide the model uuid on deploy (publish, promote)

DanielArndt:


nsklikas (approved):

DanielArndt:
> Don't you need to also update the tf plan for this to work properly?

I don't know. I just saw the `MODEL_UUID` is missing from the upstream workflow. 

I noticed #577 made some small changes, that look like they were intended to complete this, but I don't think it's enough.

I've included what I thought might need to be done, but I'll need to touch base with someone tomorrow to understand how we want this deployment stuff to work, as right now I'm not sure what the intention is.

PR #593, opened 2026-01-20
Add custom email template for verification via code + enable verification via config

BarcoMasile:
## Description
This PR:
- adds `verification_enabled` config
- updates the `kratos-info` and `login-ui-info` interfaces to include the new configs
- adds config for sender email and name
- adds support for customizing the "verification via code" email template. The file is passed to the charm as a `resource` of type `file`.
  To define the verification template, run `juju attach-resource kratos verification-email-template=./verification-email-template.gotmpl`. This will trigger a refresh event and attach the template to kratos.


src/integrations.py
```diff
@@ -250,6 +252,8 @@ def to_env_vars(self) -> EnvVars:
 
         return {
             "COURIER_SMTP_CONNECTION_URI": f"{scheme}://{self.username}:{self.password}@{self.server}:{self.port}/{option}",
+            "COURIER_SMTP_FROM_ADDRESS": "identity@canonical.com",
+            "COURIER_SMTP_FROM_NAME": "Canonical Identity Platform",
```
BarcoMasile: these should be made into config keys

src/constants.py
```diff
@@ -11,6 +11,10 @@
 PEBBLE_READY_CHECK_NAME = "ready"
 EMAIL_TEMPLATE_FILE_PATH = Path("/etc/config/templates") / "recovery-body.html.gotmpl"
 MAPPERS_LOCAL_DIR_PATH = Path("claim_mappers")
+VERIFICATION_EMAIL_TEMPLATE_RESOURCE_NAME = "verification-email-template"
+VERIFICATION_EMAIL_TEMPLATE_TARGET_PATH = Path(
```
BarcoMasile: This is used as a string in charm.py line 1148. I think we can store it directly as a string since we don't do path operations on it

BarcoMasile (commented):
LGTM (approved, but since I opened it I cannot formally approve it)

src/charm.py
```diff
@@ -1117,6 +1127,33 @@ def _get_active_flags(self) -> FeatureFlags:
 
         return feature_flags
 
+    def _push_verification_email_template(self) -> None:
```
wood-push-melon, line 1128: A suggestion (not forced): This private method is doing much. The first half seems to belong to `configs.py`, and the second half of the logic belongs to `services.py`.

src/charm.py
```diff
@@ -1117,6 +1127,33 @@ def _get_active_flags(self) -> FeatureFlags:
...
+        try:
+            content = resource_path.read_text(encoding="utf-8")
+            if content:
+                self._container.push(
+                    str(VERIFICATION_EMAIL_TEMPLATE_TARGET_PATH), content, make_dirs=True
+                )
+            else:
+                logger.debug("Provided verification email template is empty, default one will be used.")
+        except Exception as e:
+            logger.debug(
```
wood-push-melon: A nit: might be `logger.error`?

src/charm.py
```diff
@@ -584,6 +590,8 @@ def _on_upgrade_charm(self, event: UpgradeCharmEvent) -> None:
             app_name=self.app.name,
         )
 
+        self._holistic_handler(event)
```
shipperizer, line 591: how did we miss this?good catch anyway

src/constants.py
shipperizer: let's change the name then

src/charm.py
natalian98: We observe the update status event so the holistic handler was run but with some delay

wood-push-melon (approved):

PR #643, opened 2026-04-20
Add login hook integration

nsklikas:
- add login hook lib/logic
- enahce registration hook config
- fix project config

### DIFF
--- a/lib/charms/kratos/v0/kratos_login_webhook.py
+++ b/lib/charms/kratos/v0/kratos_login_webhook.py
@@ -1,0 +1,231 @@
+#!/usr/bin/env python3
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Interface library for configuring a kratos login webhook.
+
+The provider side is responsible for providing the configuration that kratos
+will use to call this webhook.
+
+The requirer side (kratos) takes the configuration provided and updates its
+config.
+"""
+
+import logging
+from functools import cached_property
+from string import Template
+from typing import Annotated, Literal, Optional
+
+from ops import (
+    CharmBase,
+    EventSource,
+    ModelError,
+    Object,
+    ObjectEvents,
+    Relation,
+    RelationBrokenEvent,
+    RelationCreatedEvent,
+    RelationEvent,
+    Secret,
+    SecretNotFoundError,
+)
+from pydantic import (
+    BaseModel,
+    BeforeValidator,
+    Field,
+    PlainSerializer,
+    StrictBool,
+    field_serializer,
+)
+
+LIBID = "070d0bcbc42042309cc556b43f3f77fd"
+LIBAPI = 0
+LIBPATCH = 1
+
+PYDEPS = ["pydantic"]
+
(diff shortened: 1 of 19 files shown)


lib/charms/kratos/v0/kratos_login_webhook.py
```diff
@@ -0,0 +1,312 @@
...
+
+        self._charm = charm
+        self._relation_name = relation_name
+
+        events = self._charm.on[relation_name]
+        self.framework.observe(events.relation_created, self._on_relation_created)
+        self.framework.observe(events.relation_broken, self._on_relation_broken)
+
+    def _on_relation_created(self, event: RelationCreatedEvent) -> None:
+        self.on.ready.emit(event.relation)
```
natalian98, line 177: suggestion: iirc the relation data is not written instantly on `relation-created`. Perhaps `relation-changed` is more suitable?

nsklikas: This is just to notify the charm that the relation is there. The charm writes the information in the databag asynchronously. The other side of the relation does not write anything in the databag. What is the issue here?

natalian98: I understood that the relation data must be available for kratos to reconfigure when the ready event is emitted. If it just need to be aware of relation presence then we can keep it as is.

natalian98 (approved):

templates/kratos.yaml.j2
```diff
@@ -33,16 +33,101 @@ identity:
...
+            {%- set _need_login_code = enable_local_idp and (_login_all or _login_after.code | d([])) %}
+            {%- set _need_login_webauthn = (enable_passwordless_login_method or enable_oidc_webauthn_sequencing) and (_login_all or _login_after.webauthn | d([])) %}
+            {%- set _need_login_totp = (enforce_mfa and enable_local_idp) and (_login_all or _login_after.totp | d([])) %}
+            {%- set _need_login_lookup = (enable_oidc_webauthn_sequencing or (enforce_mfa and enable_local_idp)) and (_login_all or _login_after.lookup_secret | d([])) %}
+            {%- if _need_login_password or _need_login_oidc or _need_login_code or _need_login_webauthn or _need_login_totp or _need_login_lookup %}
+            after:
+                {%- if _need_login_password %}
+                password:
+                    hooks:
+                        {{- render_web_hooks(_login_all, "                        ") }}
```
shipperizer, line 97: is the `"                           "` needed?what are the chances that a linter/formatter will can the spaces?

terraform/main.tf
```diff
@@ -5,11 +5,20 @@
  * using the Juju Terraform provider.
  */
 
+locals {
+  charmcraft = yamldecode(file("${path.module}/../charmcraft.yaml"))
+  oci_image = {
+    "oci-image" : local.charmcraft.resources.oci-image.upstream-source
+  }
+  resources = merge(local.oci_image, var.resources)
+}
```
shipperizer, line 14: this has issues with some of the version of the tf provider, probably better to set these as variable defaults

PR #660, opened 2026-05-15
feat: replace `traefik-route` relation with `istio-ingress-route`

natalian98:
This PR replaces the `traefik-route` relations with `istio-ingress-route`.


natalian98:
I'll keep this in draft until login ui implements the istio integration as well, otherwise we'd need to continue deploying traefik in addition to istio to provide the integration (see [here](https://github.com/canonical/kratos-operator/pull/660/changes#diff-08291d7f4ad7b2caecacb6ae26af540ebbaede9fe42d5e8e1b39d2bdefe4c2d4R99-R100))

src/charm.py
```diff
@@ -747,69 +726,24 @@ def _on_database_relation_broken(self, event: RelationBrokenEvent) -> None:
...
-        self.unit.status = MaintenanceStatus("Configuring resources")
-
         if self.unit.is_leader():
-            logger.info("This app no longer has public-route")
-
-        # needed due to how traefik_route lib is handling the event
-        self.public_route._relation = event.relation
+            route_data = PublicRouteData.load(self.public_ingress_route)
+            if route_data.config is not None:
+                self.public_ingress_route.submit_config(route_data.config)
```
nsklikas: nit (feel free to ignore): Since we are refactoring the logic, maybe we should also move this to the holistic handler.

src/integrations.py
```diff
@@ -273,57 +284,183 @@ def load(cls, requirer: SmtpRequires) -> Self:
...
+                backends=[backend],
+                matches=[
+                    HTTPRouteMatch(
+                        path=HTTPPathMatch(
+                            type=HTTPPathMatchType.PathPrefix,
+                            value="/self-service/methods/oidc/callback",
+                        )
+                    )
+                ],
+            ),
```
nsklikas, line 317: question: since we are just matching and `matches` expects a list, do we need 3 separate `HTTPRoute`?Not sure if there is a real difference between the 2. e.g.:
```
            HTTPRoute(
                name="kratos",
                listener=listener,
                backends=[backend],
                matches=[
                    HTTPRouteMatch(
                        path=HTTPPathMatch(
                            type=HTTPPathMatchType.PathPrefix,
                            value="/self-service/methods/oidc/callback",
                        )
                    ),
                    HTTPRouteMatch(
                        path=HTTPPathMatch(type=HTTPPathMatchType.PathPrefix, value="/schemas")
                        )
                    ),
                    HTTPRouteMatch(
                        path=HTTPPathMatch(
                            type=HTTPPathMatchType.PathPrefix, value="/self-service"
                        )
                    )
                ],
            ),
```

src/integrations.py
```diff
@@ -273,57 +284,183 @@ def load(cls, requirer: SmtpRequires) -> Self:
...
+        relation = requirer._charm.model.get_relation(requirer._relation_name)
+        if not relation or not relation.app:
+            return cls()
 
+        external_host = relation.data[relation.app].get("external_host", "")
         if not external_host:
             logger.error("External hostname is not set on the ingress provider")
             return cls()
 
+        tls_enabled = relation.data[relation.app].get("tls_enabled", "False") == "True"
```
nsklikas: nit (feel free to ignore): so the lib exposes properties for `external_host` and `tls_enabled` BUT they update the stored state every time and then read from it. Also afaict they only return the value if the unit is the leader (as the stored state is not shared across units). I will open an issue in the lib, as this does not really make sense. Should we add a comment to make it clear why we read directly from the databag instead of using the requirer class properties to avoid future confusion?

nsklikas: https://github.com/canonical/istio-ingress-k8s-operator/issues/171

nsklikas (approved):
lgtm, only some nits

src/integrations.py
nsklikas: The lib needs to be installed from charmlibs.

https://github.com/canonical/istio-ingress-k8s-operator/issues/171#issuecomment-4614999858

tests/integration/test_charm.py
```diff
@@ -97,18 +111,18 @@ def test_build_and_deploy(juju: jubilant.Juju, local_charm: Path) -> None:
...
+            ISTIO_INTERNAL_APP,
             LOGIN_UI_APP,
         ),
         error=any_error(
             KRATOS_APP,
             DB_APP,
-            TRAEFIK_PUBLIC_APP,
-            TRAEFIK_ADMIN_APP,
             LOGIN_UI_APP,
         ),
```
natalian98, line 124: it seems that istio often goes into error state casuing the tests to fail, so I removed it from the `any_error` check
```
'istio-internal': AppStatus(
      charm='istio-ingress-k8s',
      charm_origin='charmhub',
      charm_name='istio-ingress-k8s',
      charm_rev=61,
      exposed=False,
      base=FormattedBase(name='ubuntu', channel='24.04'),
      charm_channel='2/stable',
      charm_version='5df611c',
      scale=1,
      provider_id='62d943f3-757f-4f72-b27b-dd820898e9e7',
      address='10.152.183.141',
      app_status=StatusInfo(current='error', message='hook failed: "leader-elected"', since='23 Jun 2026 11:20:45+02:00'),
      relations={
        'istio-ingress-config': [
          AppStatusRelation(related_app='istio', interface='istio_ingress_config', scope='global'),
        ],
        'istio-ingress-route': [
          AppStatusRelation(related_app='kratos', interface='istio_ingress_route', scope='global'),
        ],
        'peers': [
          AppStatusRelation(related_app='istio-internal', interface='istio_ingress_k8s_peers', scope='global'),
        ],
      },
      units={
        'istio-internal/0': UnitStatus(
          workload_status=StatusInfo(current='error', message='hook failed: "leader-elected"', since='23 Jun 2026 11:20:45+02:00'),
          juju_status=StatusInfo(current='idle', since='23 Jun 2026 11:20:45+02:00', version='3.6.24'),
          leader=True,
          address='10.1.141.225',
          provider_id='istio-internal-0',
        ),
      },
```

BarcoMasile (approved):

PR #666, opened 2026-06-08
Add leader guard to secret_remove

nsklikas:
For non-leader units the `secret_get` would succeed (as all units can read a secret), but then then the `remove_all_revisions` would raise a `SecretNotFoundError` (as non-leader units do not have permissions to edit the secret).


natalian98 (dismissed):

BarcoMasile (dismissed):

natalian98:
`charmcraft pack` fails due to https://github.com/canonical/charmcraft/issues/2720

natalian98 (approved):

PR #683, opened 2026-06-30
test: adopt -Werror in unit tests

tonyandrewmeyer:
Add [tool.pytest.ini_options] filterwarnings = ["error", ...] so the test suite fails on deprecations and resource leaks, with three narrow, message-anchored ignores:

* Harness is deprecated; ignored until tests/unit/test_charm.py moves to Scenario.
* The vendored data-interfaces lib gets JujuVersion from the environment rather than the model; ignored until fixed in the vendored lib.
* Integration tests hit the deployed charm over HTTPS with self-signed certs and verify=False; urllib3's InsecureRequestWarning is ignored as a by-design signal.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,6 +12,21 @@ show_missing = true
 minversion = "6.0"
 log_cli_level = "INFO"
 pythonpath = ["src", "lib", "tests"]
+filterwarnings = [
+    "error",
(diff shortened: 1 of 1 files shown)


natalian98 (approved):
thanks for the contribution @tonyandrewmeyer!
I rebased your branch, it should be good to go after the CI passes.

PR #702, opened 2026-07-17
refactor: migrate to uv

natalian98:


BarcoMasile (dismissed):

wood-push-melon (dismissed):

charmcraft.yaml
natalian98, line 1: I changed `charmcraft.yaml`: with uv the schemas and claim mappers were nested into `schemas/identity_schemas/identity_schemas/` causing the tests to fail, now they're at the root as expected by the charm (`identity_schemas/` and `claim_mappers/`)

